PR #611, opened 2014-11-26
Thesis orientator restritions

diutsu:
...and email signatures for thesis jury approval from vice president of scientific council


pedrosan7os:
"Prevents thesis orientator from being the jury president" commit is not a universal restriction and should not be integrated, remove this from the commit list

PR #636, opened 2015-01-21
multiple changes

NunoPinheiro:


pedrosan7os:
This way is impossible to validade or discuss, please separate into several pull requests, each related with a specific JIRA issue.

PR #643, opened 2015-01-21
closes ACADEMIC-636

NunoPinheiro:


src/main/java/org/fenixedu/academic/domain/Degree.java
```diff
@@ -316,6 +316,30 @@ public DegreeCurricularPlan createPreBolonhaDegreeCurricularPlan(String name, De
         }
     }
 
+    @Deprecated
+    public DegreeCurricularPlan createDegreeCurricularPlan(String name, GradeScale gradeScale, Person creator) {
```
jcarvalho: What is the point on creating a new already-deprecated method in a core domain class, when its sole purpose is to be used in a service?

NunoPinheiro: wrong method, will review tomorrow

PR #644, opened 2015-01-22
Bologna independent DCP group structure creation

NunoPinheiro:
Closes [ACADEMIC-636](https://jira.fenixedu.org/browse/ACADEMIC-636)

### DIFF
--- a/src/main/java/org/fenixedu/academic/domain/Degree.java
+++ b/src/main/java/org/fenixedu/academic/domain/Degree.java
@@ -316,27 +316,32 @@ public class Degree extends Degree_Base implements Comparable<Degree> {
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/fenixedu/academic/domain/Degree.java
```diff
@@ -316,6 +316,29 @@ public DegreeCurricularPlan createPreBolonhaDegreeCurricularPlan(String name, De
         }
     }
 
+    public DegreeCurricularPlan createDegreeCurricularPlan(String name, GradeScale gradeScale, Person creator) {
```
jcarvalho: With this change, won't the code in `createBolonhaDegreeCurricularPlan` be duplicated here? If so, in `createBolonhaDegreeCurricularPlan`, in the `isBolonhaDegree` case, just invoke this method.

PR #646, opened 2015-01-23
emiting signal on enrolments. removed scientific comission verification on Thesis constructor

pedrorijo91:


pedrosan7os:
squash your commits, also avoid formatting javadoc

PR #647, opened 2015-01-29
Forces the use of 3 letter country code

cfscosta:
The 3 letter country code is now verified correctly and produces the desired BootstrapError f an incorrect code is sent. Be it an incorrect one or a two letter country code.

Fixes: https://jira.fenixedu.org/browse/ACADEMIC-333


nurv:
this should change into a list of countries, but for now this works.

PR #648, opened 2015-01-29
API enhancement to improve legacy data migration activities in any given context

NunoPinheiro:
closes https://jira.fenixedu.org/browse/ACADEMIC-637


src/main/java/org/fenixedu/academic/domain/DegreeCurricularPlan.java
```diff
@@ -1864,6 +1865,12 @@ public boolean canSubmitImprovementMarkSheets(final ExecutionYear executionYear)
         return sortedExecutionDegrees.last().getExecutionYear().equals(executionYear.getPreviousExecutionYear());
     }
 
+    public ExecutionInterval getBegin() {
+        final List<ExecutionYear> executionYears = new ArrayList<ExecutionYear>(getBeginContextExecutionYears());
```
jcarvalho: This seems like a really wasteful way of doing this. You should use [Collections.min](http://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#min-java.util.Collection-java.util.Comparator-) or something similar instead.

src/main/java/org/fenixedu/academic/domain/DegreeCurricularPlan.java
```diff
@@ -1864,6 +1865,10 @@ public boolean canSubmitImprovementMarkSheets(final ExecutionYear executionYear)
         return sortedExecutionDegrees.last().getExecutionYear().equals(executionYear.getPreviousExecutionYear());
     }
 
+    public ExecutionInterval getBegin() {
+        return Collections.min(getBeginContextExecutionYears(), ExecutionYear.COMPARATOR_BY_YEAR);
```
jcarvalho: Beware that if the DCP has no associated execution year, this will throw a `NoSuchElementException`, instead of returning null.

PR #649, opened 2015-01-29
Added functionality to mark optional course groups.

NunoPinheiro:
closes https://jira.fenixedu.org/browse/ACADEMIC-630


src/main/java/org/fenixedu/academic/domain/degreeStructure/CourseGroup.java
```diff
@@ -842,4 +857,74 @@ public void applyToCurricularCourses(final ExecutionYear executionYear, final Pr
         }
     }
 
+    public Context findChild(final MultiLanguageString name) {
```
jcarvalho: Given the 'fuzzy' nature of this method, it should either be very well documented, or be given a name that better conveys what it does.

src/main/java/org/fenixedu/academic/service/services/bolonhaManager/EditCourseGroup.java
```diff
@@ -27,13 +27,22 @@
 import org.fenixedu.academic.service.services.exceptions.FenixServiceException;
 
 import pt.ist.fenixframework.Atomic;
+import pt.ist.fenixframework.DomainObject;
 import pt.ist.fenixframework.FenixFramework;
 
 public class EditCourseGroup {
 
-    @Atomic
```
jcarvalho: This way wouldn't the old method no longer be @Atomic? That can cause issues with other methods invoking it.

PR #654, opened 2015-02-09
Degree duration shouldn't be restricted to correspondent DegreeType years

nadiramin:
Closed ACADEMIC-623


jcarvalho:
There was an issue with the Maven Repository. It was fixed, so the Travis Build is now OK.

src/main/java/org/fenixedu/academic/domain/DegreeCurricularPlan.java
```diff
@@ -2008,4 +2031,59 @@ public void setInitialDate(java.util.Date date) {
         return getTargetEquivalencePlansSet();
     }
 
+    public int getDurationInYears() {
+        return Float.valueOf(getDegreeStructure().getAcademicPeriod().getWeight()).intValue();
```
pedrosan7os: The degreeStructure relation is not required and the existing data contains DegreeCurricularPlans without a CurricularPeriod. The relation should be required and the existing data properly filled, but that must be done in the next major. For now, all accesses to getDegreeStructure() should be safeguarded from nulls and the duration returned from the degree type as a fallback option.

PR #655, opened 2015-02-09
New interface to create Academic Offices.

diutsu:
Dependent on fenixedu/fenixedu-spaces#26. Needs the SpaceJsonAdapter to work.


src/main/java/org/fenixedu/academic/json/adapters/AdministrativeOfficeJsonAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+    public JsonElement view(AdministrativeOffice administrativeOffice, JsonBuilder ctx) {
+        JsonObject object = new JsonObject();
+        object.add("name", ctx.view(administrativeOffice.getName()));
+        object.addProperty("externalId", administrativeOffice.getExternalId());
+        object.add("space", ctx.view(administrativeOffice.getCampus()));
+        object.add("coordinator", ctx.view(administrativeOffice.getCoordinator()));
+        return object;
+    }
+
+    @Atomic
```
jcarvalho: The annotation is not needed here, as the JsonAdapter infrastructure will already provide the write transaction.

src/main/java/org/fenixedu/academic/json/adapters/DepartmentJsonAdapter.java
```diff
@@ -16,7 +16,7 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with FenixEdu Academic.  If not, see <http://www.gnu.org/licenses/>.
  */
-package org.fenixedu.academic.ui.spring.controller.manager;
+package org.fenixedu.academic.json.adapters;
```
jcarvalho: While this is the package the adapter should be defined on, changing it would be an API break.

pom.xml
```diff
@@ -245,7 +245,7 @@
         <dependency>
             <groupId>org.fenixedu</groupId>
             <artifactId>fenixedu-spaces</artifactId>
-            <version>1.3.1</version>
+            <version>1.4.0-SNAPSHOT</version>
```
jcarvalho: If you REALLY depend on this new version, wait for it to be released before changing it in the POM.

PR #659, opened 2015-02-10
fixed evaluation for delegates backbean package in faces config file

pedrorijo91:


pedrorijo91:
fenixedu-ist-delegates waiting release with this fix

PR #661, opened 2015-02-11
fixed label when adding web address to personal information

pedrorijo91:


src/main/webapp/person/visualizePersonalInfo.jsp
```diff
@@ -40,7 +40,7 @@
 </h2>
 
 <logic:messagesPresent message="true">
-	<html:messages id="messages" message="true" bundle="APPLICATION_RESOURCES">
+	<html:messages id="messages" message="true" bundle="ACADEMIC_OFFICE_RESOURCES">
```
pedrosan7os: In this case it would make more sense to add the missing label in the APPLICATION_RESOURCES, (or any other of the scope of this interface), rather than having a functionality related the user's own information dependent on academic office resource bundles.

src/main/webapp/person/visualizePersonalInfo.jsp
```diff
@@ -122,9 +122,10 @@
 	</tr>
 </table>
 
+
 <logic:messagesPresent message="true" property="contacts">
 	<ul class="nobullet list6">
-		<html:messages id="messages" property="contacts" message="true">
+		<html:messages id="messages" property="contacts" message="true" bundle="ACADEMIC_OFFICE_RESOURCES">
```
pedrosan7os: same here.

PR #664, opened 2015-02-16
Adds 'all' option to filters on student list

diutsu:
Enrolment count are the total number of enrolments on the course.


diutsu:
Regarding the ordering of the columns, I haven't though how groupings and shifts columns can be ordered, but in any case the order option can be added to any of the other columns.

PR #668, opened 2015-02-18
Fixed groupings.jsp labels

ricardosequeira:


jcarvalho:
Please squash your commits into a single one, explaining the problem.

Also, please avoid making such large changes in the property files, by looking at the diff, we can't really see what is being changed.

jcarvalho:
FYI, you don't need to close the pull request to make changes, you can simply amend your commits and `push -f` to your fork. We actually prefer you do that, so we can keep track of the whole conversation.

ricardosequeira:
I made small changes through github editor which changed several lines of files

PR #671, opened 2015-02-23
Person creation signals

mbborgez:
Signals are emmited when a person is created


src/main/java/org/fenixedu/academic/domain/Person.java
```diff
@@ -119,7 +121,7 @@
 public class Person extends Person_Base {
 
     private static final Integer MAX_VALIDATION_REQUESTS = 5;
-
+    public static final String CREATED_SIGNAL = "academic.person.create.signal";
```
pedrosan7os: Change this to: ```public static final String CREATED_SIGNAL = "fenixedu.academic.person.create";```

PR #673, opened 2015-02-24
Fixes on student group interface

diutsu:
- Fix: Removes minimum number of elements on individual groupings.
- Fix: Removes persistent error messsages
- Fix: Enables deselection of grouping elements when group has maximum size
- Fix: Corrects behaviour of  clear selection
- Fix: Labels

### DIFF
--- a/src/main/webapp/student/groupings/studentGroup.jsp
+++ b/src/main/webapp/student/groupings/studentGroup.jsp
@@ -142,7 +142,6 @@
 			</tr>
 		</tbody>
 	</table>
-
 	<div class="modal fade" id="editShift" tabindex="-1" role="dialog"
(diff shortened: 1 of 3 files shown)


src/main/webapp/student/groupings/studentGroup.jsp
```diff
@@ -30,7 +30,7 @@
 		class="glyphicon glyphicon-chevron-left"></span>
 		${fr:message('resources.ApplicationResources', 'label.back')} </a>
 	<div class="alert alert-info" ng-show="message">{{ message }}</div>
-	<div class="alert alert-danger" ng-show="error">{{ error }}</div>
+	<div class="alert alert-danger" ng-show="error">{{ error }}</span></div>
```
pedrosan7os: where is this span being open, are you sure this is correct?

PR #685, opened 2015-03-14
Recover interfaces for coordinators to edit degree information. ACDM-661 #resolve

mbborgez:
interfaces for coordinators to mange degree information was removed on last version and was now recovered with some minor improvements and adaptations.

ACDM-661 #resolve


pedrosan7os:
JIRA commit directives must go in the title, as comments they won't work

src/main/webapp/WEB-INF/resources/ApplicationResources_pt.properties
```diff
@@ -14,7 +14,7 @@ label.delete = Apagar
 label.department = Departamento
 label.edit = Editar
 label.listing = Listar
-label.manageDepartments = Gest�o de Departamentos
+label.manageDepartments = Gest�o de Departamentos
```
pedrosan7os: Watch out for editors that alter the encoding

mbborgez: Thanks. It has been corrected.

PR #686, opened 2015-03-16
allowed multiple Advisors in the same Thesis #resolve ACDM-668

pedrorijo91:


pedrosan7os:
This is not sufficient to solve the orientation problem. There is a lot of behaviour biased to the fact that thesis have 1 orientator and 1 co-orientator

pedrorijo91:
This modification only allows that thesis proposals (which may have multiple Advisors/Orientadores) may be converted into Thesis objects

pedrorijo91:
Fully solved in https://github.com/FenixEdu/fenixedu-academic/pull/693/files#diff-61a54d7481920fcf7b1ea1eb0056f1cb

PR #688, opened 2015-03-16
Allows Scientific Council to insert thesis files from stage E. Fixes bugs concerning thesis declaration. ACDM-688 #resolve

gil-l:
Allows Scientific Council to insert thesis files from stage E. Fixes bugs concerning thesis declaration.

Scientific Council can now insert thesis files (and not just replacing existing ones) starting from stage E
of the thesis process. Their visibility is set as Intranet and the user is informed if the student has not yet
accepted the thesis declaration. Also fixes the visibility presentation on the thesis declaration page 
and the JSP error on said page when student rejects and has uploaded thesis files.

IST RT #446608


gil-l:
roger ;)

PR #689, opened 2015-03-17
Refactored associated objects interface

diutsu:


diutsu:
rebased on master

pedrosan7os:
- The labels have bad encoding
- The departments tab is empty for a database full of departments, is that normal?
- When creating a department the 'active' is an option. That is too CRUD for me, if the semantic is to support like draft -> published behaviour the interface should give those actions instead of a checkbox field.
- The Scientific Areas tab should group more, this is a flat list with hundreds of elements
- 'Coordinator Username' should be 'Coordinator'

diutsu:
Everythings fixed now.

Regarding:
    The Scientific Areas tab should group more, this is a flat list with hundreds of elements

There is a filter on the page to filter and also on the department interface there is a button to view only the scientific areas of a specific department

jcarvalho:
It seems your editor has greatly mangled the properties files, please minimize the changeset to the absolutely necessary.

jcarvalho:
Some remarks:

- Despite the domain containing separate slots for the department's name in Portuguese and English, the UI should show a LocalizedString.
- The `Active` button just looks awkward.
- Why do some pages have breadcrumbs and others do not?
- Given the strong relation between Departments and Scientific Areas, and that you can see the areas of a given department, I'm not sure it makes sense to have a whole tab dedicated to it.
- No access control is being made on any of the REST endpoints.
- You can use @RestController, instead of annotating all methods in the controller with @ResponseBody.
- The functionality's English name is 'School Structure Management65', and some English labels are missing (e.g. `!label.competenceCourse!`).
- Some interfaces are missing a 'Loading' message.
- How is the Empty Degree managed now?

jcarvalho:
- In the Administrative Offices tab, clicking on 'Assign Authorization Management', the confirmation modal shows the following text: `!message.manage.authorizations.self.warning!`
- The inner breadcrumbs are still messy, wouldn't it be better if the relevant pages just contained a proper title?
- The 'Create Department' button has the following label: `Create !label.manager.department!`
- In the departments view, the Real Name label is allows prefixed by `pt-PT - pt`
- Why do departments have two distinct 'Acronym's?
- If the Department's Real Name does not end with `(<ACRONYM>)`, the Create operation throws an exception and silently fails.
- When creating a Department with a repeated Acronym, the following error is shown `!error.unit.already.exists.unit.with.same.name.or.acronym!`.
- If the Department's Real Name is not filled in English, an error is thrown.
- The `Active` button must really be a checkbox.
- The title for the Competence Course interface is `!label.competenceCourse!`.
- The Competence Course acronym seems to be optional, even though the interface is requiring it.
- The `competence courses` title is presented in lowercase in many locations.
- When showing the Scientific Areas, it should be clear the department they are inserted in.
- The Scientific Areas page seems a bit too gray...

pedrosan7os:
This a good starting point, but not ready for integration.

PR #693, opened 2015-03-23
allowing multiple Orientator in the same Thesis due to legal requirements ACDM-668 #resolve

pedrorijo91:


sergiofbsilva:
please rebase this since the merge fails with conflicts.

src/main/java/org/fenixedu/academic/domain/thesis/ThesisEvaluationParticipant.java
```diff
@@ -68,11 +68,37 @@ public double getParticipationCredits() {
...
+                if (thesis.getCoorientatorCreditsDistribution() != null) {
+                    return thesis.getCoorientatorCreditsDistribution();
+                }
+            }
+        }
+
+        return 0;
+    }
+
+    public Integer getCreditsDistribution() {
```
pedrosan7os: type cannot change to Integer because it breaks the API

PR #704, opened 2015-04-02
Signal raising logic for Degree creation.

sanavarali:
Relates to #LRN-20
(after the acceptance and release of this request, there will be made a commit in fenixedu-learning that depends on this)


pedrosan7os:
Assigned issue ACDM-702

PR #708, opened 2015-04-13
Added option to expand course groups when dealing with large degree curricular plans

nadiramin:
Closes ACDM-712


jcarvalho:
Shouldn't the JSP files in the Scientific Council and Bolonha Portals be included as well? Or is the issue specific to this one interface?

PR #710, opened 2015-04-14
Changed group presentation names to be more uniform and readable ACDM-715 #resolve

gil-l:
All groups now have a localizable names. The names display all member
filter attributes present in the group. These have the same
representation across all group classes. Portuguese names were made to
sound more natural. Corrected other name inconsistencies across groups
through the property bundles. Added some keys for group names in FIST
modules that still use academic bundles. Additionally cleaned up related
unused keys in the group name bundle and others.

Related Jira Issues: ACDM-715, FIST-86


jcarvalho:
Please rebase this pull request, as some incompatible changes have been introduced in `master`.

PR #712, opened 2015-04-16
[Student] Refactor: Created a StatuteType domain entity to replace the old enum StudentStatuteType.

nadiramin:
Closes ACADEMIC-648.

Note: Migration guide was updated but migration scripts were not tested since we don't have a working database.


src/main/webapp/WEB-INF/academic-configurations/statutetypemanagement/statutetype/update.jsp
```diff
@@ -0,0 +1,222 @@
...
+<spring:url var="datatablesI18NUrl" value="/javaScript/dataTables/media/i18n/${portal.locale.language}.json"/>
+
+<link rel="stylesheet" type="text/css"
+	href="${pageContext.request.contextPath}/CSS/dataTables/dataTables.bootstrap.min.css"/>
+
+<link href="//cdn.datatables.net/responsive/1.0.4/css/dataTables.responsive.css" rel="stylesheet"/>
+<script src="//cdn.datatables.net/responsive/1.0.4/js/dataTables.responsive.js"></script>
+<link href="//cdn.datatables.net/tabletools/2.2.3/css/dataTables.tableTools.css" rel="stylesheet"/>
+<script src="//cdn.datatables.net/tabletools/2.2.3/js/dataTables.tableTools.min.js"></script>
+<link href="//cdnjs.cloudflare.com/ajax/libs/select2/4.0.0-rc.1/css/select2.min.css" rel="stylesheet" />
```
jcarvalho: Please don't add external resource links, specially to libraries in Release Candidate :)

src/main/java/org/fenixedu/academic/ui/spring/controller/StatuteTypeManagement/StatuteTypeController.java
```diff
@@ -0,0 +1,190 @@
...
+import pt.ist.fenixframework.Atomic;
+
+@SpringFunctionality(app = AcademicConfigurationsController.class, title = "label.title.StatuteTypeManagement",
+        accessGroup = "#managers")
+@RequestMapping("/academic-configurations/statutetypemanagement/statutetype")
+public class StatuteTypeController extends AcademicBaseController {
+    @RequestMapping
+    public String home(Model model) {
+        //this is the default behaviour, for handling in a Spring Functionality
+        return "redirect:/academic-configurations/statutetypemanagement/statutetype/";
```
jcarvalho: Why this redirect?!

src/main/java/org/fenixedu/academic/ui/spring/controller/StatuteTypeManagement/StatuteTypeController.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @RequestMapping(value = "/read/{oid}")
+    public String read(@PathVariable("oid") StatuteType statuteType, Model model) {
+        setStatuteType(statuteType, model);
+        return "academic-configurations/statutetypemanagement/statutetype/read";
+    }
+
+    @RequestMapping(value = "/delete/{oid}")
+    public String delete(@PathVariable("oid") StatuteType statuteType, Model model) {
```
jcarvalho: This seems like too much of a delicate operation to be performed in a GET request.

src/main/dml/fenixedu-academic.dml
```diff
@@ -3016,9 +3015,25 @@ valueType org.fenixedu.academic.domain.degreeStructure.CycleTypes as CycleTypes
     	class student.StudentDataShareStudentsAssociationAuthorization extends student.StudentDataShareAuthorization {
     	}
 
+	class student.StatuteType {
+		String code;
+		LocalizedString name;
+		Boolean grantsWorkingStudentStatute;
```
jcarvalho: Why not make these primitive `boolean`, and avoid the null-checks all around?

src/main/java/org/fenixedu/academic/domain/student/StatuteType.java
```diff
@@ -0,0 +1,200 @@
...
+        check(getName(), "error.StatuteType.name.required");
+    }
+
+    private static void check(final Object obj, final String message, final String... args) {
+        if (obj == null) {
+            throw new DomainException(message, args);
+        }
+    }
+
+    public boolean isGrantsWorkingStudentStatute() {
```
jcarvalho: The name of this method sounds somewhat weird

src/main/java/org/fenixedu/academic/domain/student/StatuteType.java
```diff
@@ -0,0 +1,200 @@
...
+        return CollectionUtils.select(StatuteType.readAll(), new Predicate() {
+
+            @Override
+            public boolean evaluate(Object arg0) {
+                return ((StatuteType) arg0).isActive();
+            }
+        });
+    }
+
+    public static StatuteType find(String code) {
```
jcarvalho: All these `find` methods should return an `Optional`.

src/main/java/org/fenixedu/academic/domain/student/StatuteType.java
```diff
@@ -0,0 +1,200 @@
...
+import org.fenixedu.academic.domain.DomainObjectUtil;
+import org.fenixedu.academic.domain.exceptions.DomainException;
+import org.fenixedu.bennu.core.domain.Bennu;
+import org.fenixedu.commons.i18n.LocalizedString;
+
+import pt.ist.fenixframework.Atomic;
+
+public class StatuteType extends StatuteType_Base {
+
+    public static Comparator<StatuteType> COMPARATOR_BY_NAME = new Comparator<StatuteType>() {
```
jcarvalho: You could do this simply with:
```java
Comparator.comparing(StatuteType::getName).thenComparing(DomainObjectUtil.COMPARATOR_BY_ID);
```

src/main/java/org/fenixedu/academic/domain/student/StatuteType.java
```diff
@@ -0,0 +1,200 @@
...
+    public void delete() {
+        if (!isDeletable()) {
+            throw new DomainException("error.StatuteType.deletion.not.possible");
+        }
+
+        setRootDomainObject(null);
+        deleteDomainObject();
+    }
+
+    public static Collection<StatuteType> readAll() {
```
jcarvalho: This should just return a `Stream`, as it is mostly used to filter and lookup values.

src/main/java/org/fenixedu/academic/dto/student/StudentStatuteBean.java
```diff
@@ -84,10 +83,8 @@ public String getEndPeriodFormatted() {
     }
 
     public String getDescription() {
-        return BundleUtil.getString(Bundle.ENUMERATION, getStatuteType().getDeclaringClass().getSimpleName() + "."
-                + getStatuteType().name())
+        return getStatuteType().getName().getContent(I18N.getLocale())
```
jcarvalho: Beware that invoking [`getContent(I18N.getLocale())`](https://github.com/FenixEdu/fenixedu-commons/blob/v1.1.0/src/main/java/org/fenixedu/commons/i18n/LocalizedString.java#L461-L475) is different than just invoking [`getContent()`](https://github.com/FenixEdu/fenixedu-commons/blob/v1.1.0/src/main/java/org/fenixedu/commons/i18n/LocalizedString.java#L480-L496).

src/main/java/org/fenixedu/academic/ui/spring/AcademicBaseController.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.List;
+
+import javax.servlet.http.HttpServletRequest;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.ModelAttribute;
+
+public class AcademicBaseController {
+    //The HTTP Request that can be used internally in the controller
```
jcarvalho: As you're only using this in one controller, just put this implementation there, and it will eventually be pulled up in the future, if necessary.

[YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

src/main/java/org/fenixedu/academic/ui/spring/AcademicBaseController.java
```diff
@@ -0,0 +1,39 @@
...
+    protected @Autowired HttpServletRequest request;
```
jcarvalho: Good Spring practices dictate that access to the raw request should be kept to a minimum, so it is bad practice to expose it by default in every situation.

src/main/java/org/fenixedu/academic/ui/spring/AcademicBaseController.java
NunoPinheiro: After this pull request being accepted, another will be sent using this.

src/main/java/org/fenixedu/academic/ui/spring/AcademicConfigurationsController.java
```diff
@@ -0,0 +1,11 @@
+package org.fenixedu.academic.ui.spring;
+
+import org.fenixedu.bennu.spring.portal.SpringApplication;
+import org.fenixedu.bennu.spring.portal.SpringFunctionality;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@RequestMapping("/academic-configurations")
+@SpringApplication(group = "logged", path = "academic-configurations", title = "title.academic.configurations")
+@SpringFunctionality(app = AcademicConfigurationsController.class, title = "title.academic.configurations")
+public class AcademicConfigurationsController {
```
jcarvalho: What is the purpose of this? (I can understand the class being marked as an Application, but not the marking it as a Controller).

src/main/java/org/fenixedu/academic/ui/spring/controller/StatuteTypeManagement/StatuteTypeController.java
```diff
@@ -0,0 +1,190 @@
+package org.fenixedu.academic.ui.spring.controller.StatuteTypeManagement;
```
jcarvalho: This package name does not adhere to conventions.

src/main/java/org/fenixedu/academic/ui/spring/controller/StatuteTypeManagement/StatuteTypeController.java
```diff
@@ -0,0 +1,190 @@
...
+        accessGroup = "#managers")
+@RequestMapping("/academic-configurations/statutetypemanagement/statutetype")
+public class StatuteTypeController extends AcademicBaseController {
+    @RequestMapping
+    public String home(Model model) {
+        //this is the default behaviour, for handling in a Spring Functionality
+        return "redirect:/academic-configurations/statutetypemanagement/statutetype/";
+    }
+
+    private StatuteType getStatuteType(Model m) {
```
jcarvalho: Why are you accessing this directly?!

src/main/java/org/fenixedu/academic/ui/spring/AcademicBaseController.java
jcarvalho: Nevertheless, I do not believe this 'functionality' (which is provided by Spring in a much better way) justifies creating this super-class.

I don't want another `FenixDispatchAction`, perpetuating bad practices, simply for the reason they are placed on a base class.

NunoPinheiro: Could you provide a reference to the Spring feature which covers this behavior?
I think we could refactor it to be done through dependency injection.
As long as I don't have to replicate code in different classes, any suggestion will be considered

jcarvalho: http://docs.spring.io/spring/docs/current/spring-framework-reference/html/mvc.html

Lookup the sections about form error handling.

doc/migration/4.x-5.0.md
```diff
@@ -6,9 +6,11 @@
 
 2. Run the following script to export the currently available EvaluationSeasons: [DumpEvaluationSeason](https://gist.github.com/pedrosan7os/53ff0165cf13266857f9). Keep the generated file.
 
+3. Run the following script to export the currently available StatuteTypes: [StatuteTypes](https://gist.github.com/NunoPinheiro/9fd5bb667e8b1d77de03). Keep the generated file.
```
pedrosan7os: The generated sql has syntax errors, json parts like {"pt-PT":"Finalista"} should be around '

pedrosan7os: We improved scripts like this to avoid possible collisions of the next domain class id. I've ported that change to your script and put in my fork of your gist script. Feel free to use that change.

src/main/java/org/fenixedu/academic/ui/struts/action/administrativeOffice/lists/StudentListByDegreeDA.java
```diff
@@ -350,8 +351,8 @@ private void fillSpreadSheetFilters(SearchStudentsByDegreeParametersBean searchB
         spreadsheet.newHeaderRow();
         if (searchBean.hasAnyStudentStatuteType()) {
             spreadsheet.addHeader(getResourceMessage("label.statutes") + ":");
-            for (StudentStatuteType statute : searchBean.getStudentStatuteTypes()) {
-                spreadsheet.addHeader(BundleUtil.getString(Bundle.ENUMERATION, statute.name()));
+            for (StatuteType statute : searchBean.getStudentStatuteTypes()) {
+                spreadsheet.addHeader(statute.getName().getContent(I18N.getLocale()));
```
pedrosan7os: getContent() is the common practice to show localized content

src/main/java/org/fenixedu/academic/ui/spring/AcademicBaseController.java
```diff
@@ -0,0 +1,69 @@
+package org.fenixedu.academic.ui.spring;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.ModelAttribute;
+import org.springframework.web.servlet.mvc.support.RedirectAttributes;
+
+public class AcademicBaseController {
```
pedrosan7os: A pull request of a feature should not add new conventions, should just stick to the existing ones. Currently error handling is done directly using the spring features, and that is how it should be done until a new convention is taken.

src/main/webapp/static/academic/css/dataTables.responsive.css
```diff
@@ -0,0 +1,94 @@
+table.dataTable.dtr-inline.collapsed tbody td:first-child,
```
pedrosan7os: dataTables is already responsive via integration with bootstrap: https://github.com/FenixEdu/fenixedu-academic/blob/master/src/main/webapp/javaScript/dataTables/media/js/jquery.dataTables.bootstrap.min.js
Why are you adding another dependency?

src/main/webapp/static/academic/css/select2.min.css
```diff
@@ -0,0 +1 @@
...dius:0;}.select2-container--classic.select2-container--open.select2-container--below .select2-selection--multiple{border-bottom:none;border-bottom-left-radius:0;border-bottom-right-radius:0;}.select2-container--classic .select2-search--dropdown .select2-search__field{border:1px solid #aaa;outline:0;}.select2-container--classic .select2-search--inline .select2-search__field{outline:0;}.select2-container--classic .select2-dropdown{background-color:white;border:1px solid transparent;}.select2-container--classic .select2-dropdown--above{border-bottom:none;}.select2-container--classic .select2-dropdown--below{border-top:none;}.select2-container--classic .select2-results>.select2-results__options{max-height:200px;overflow-y:auto;}.select2-container--classic .select2-results__option[role=group]{padding:0;}.select2-container--classic .select2-results__option[aria-disabled=true]{color:grey;}.select2-container--classic .select2-results__option--highlighted[aria-selected]{background-color:#3875d7;color:white;}.select2-container--classic .select2-results__group{cursor:default;display:block;padding:6px;}.select2-container--classic.select2-container--open .select2-dropdown{border-color:#5897fb;}
```
pedrosan7os: This is a 'select' replacement, the current dependencies are sufficient, stick to using them

pedrosan7os:
As a general rule, developing a new feature should adhere to the existing javascript 'toolkit', adding new javascript is adding a dependency and should not be taken lightly. And if really really needed, should be added via webjars, and even then should be in a different pull request of an issue explaining why is that really needed.

This is not the case of this pull request, as several new libraries are added just to create the simplest CRUD interface for a statute entity.

src/main/java/org/fenixedu/academic/ui/spring/AcademicBaseController.java
ricardocb: This pull request was made following the meeting on 24/04/2015. 

While there is no common service in FenixEdu for handling the messages, the common features for adding info/warning/error messages is defined in a common code base so that it can be reused in different controllers.

Creating an abstraction for handling the messages can be used in order to pass the messages also when a redirect is made using flashattributes transparently.

src/main/webapp/static/academic/css/dataTables.responsive.css
ricardocb: This extension was included based on this documentation:
https://datatables.net/extensions/responsive/examples/styling/bootstrap.html

src/main/webapp/static/academic/css/select2.min.css
ricardocb: You're right, in this pull-request it is not needed so we will remove it (the css will also be removed for now).

But at the 24/04/2015 meeting it was agreed that the inclusion of select2 library in toolkit should be done in a near future. In the meanwhile, as agreed, it will be used as static library or webjar.

pedrosan7os: This pull request needs to use only the current javascript dependencies. There is no justifiable reason for an interface that is a simple CRUD to add a load of new dependencies.

In the discussion you refer we only stated the process of adding new dependencies, we did not agree to add anything in the near future, we just agreed to be open to discuss it. But nevertheless until it's there you can't use it. As said before adding dependencies is not something a pull requester of a feature is entitled to do without a very good reason, and I see none.

jcarvalho:
The migration guide for this branch has been updated due to the following issues:

- The provided Java script did not compile
- The generated SQL was not valid:
  - Strings were not being escaped
  - The OID column was missing
  - The number of columns did not match the number of values
  - boolean types were not being properly encoded as `0` and `1`
  - A non-existing column was being populated
  - The Working Student statute column had the wrong name
- The hard-coded Migration SQL contained a column with the wrong name
- One `UPDATE` statement for each OID was too inefficient, changed it to perform bulk updates

PR #717, opened 2015-04-20
Program conclusion configuration for course groups. ACDM-686 ACDM-690 #resolve

sergiofbsilva:
The program conclusion is now a configurable entity.
It defines its name, the graduation title (which can be parameterized with the
${degreeName} variable), if provides alumni, if the final average is
editable in the conclusion process interface and the target registration
state.
New features:
- Added interface to manage program conclusions.
- When creating or editing a course group in a degree curricular plan
the user can select a program conclusion.
- The conclusion process interface provide a list of the available
program conclusions for the current student, and the user must
select the one to conclude.
- Registry code generator does not depend on cycle type. A default
registry code generator was added with the default implementation.
If necessary for your institution, override the getCode method in
a subclass.
- Requested cycle necessary for document requests is given by
program conclusion instead of cycle type selection.
- Replace calls to degree type unimplemented API to equivalent
methods.
- Remove automatic attribution of senior statute

Issues: ACDM-686 ACDM-690


sergiofbsilva:
- Registration.hasConcluded and StudentCurricularPlan.isConclusionProcessed are not working correctly
    * RootCourseGroups are not considered, only RegistrationConclusionProcess and CycleCurriculumGroups are verified
- The registration conclusion UI should allow override of both rounded and unrounded averages
- Missing label "curriculumGroup.name"on registration conclusion detail page
- It should be possible to revert / delete a conclusion process
- The registration conclusion UI throws an error if registration does not have any approvals
- The registration conclusion document UI throws an error
    * References old degreeType (enum) attributes (administrativeOfficeType and localizedName)
- The registration conclusion UI shows a blank page when choosing the program conclusion to process
    * mapping "chooseCycleForRegistrationConclusion"in action "RegistrationDA"points to"chooseCycleForRegistrationConclusion.jsp"instead of "chooseProgramConclusion.jsp"
- RootCourseGroup should allow association of CreditsLimit rule
    * This is required for program conclusion to work correctly
- Management UI does not allow association of configurations to RootCourseGroup
    * This is needed to calculate the completion for non-Bologna degrees (without the use of RegistrationConclusionProcess) and to calculate the average of combined cycles in integrated masters (1st + 2nd cycles).
 The conclusion algorithm must be changed to include RootCurriculumGroups.
- The conclusion for integrated master degree throws an error when 2nd cycle curriculum group does not exist on student curricular plan
    * RegistrationConclusionBean assumes that all cycle (curriulum groups) supported by degree type are already created on student curricular plan, but this not always the case (newly created registrations, students who don't want to continue the 2nd cycle, affinity cycles, ...)
- Registration conclusion UI throws an error
    * registrationConclusion.jsp is testing averageEditable instead of editableAverage attribute
- ProgramConclusion.findByCode should be static
- Management UI does not save averageEditable and alumni parameters
- Management UI does not allow to delete configurations
    * Delete operation is required to correct user mistakes
- Management UI does not allow to create configurations without target registration state
    * Some program conclusions does not change state (for instance partial)

PR #720, opened 2015-04-23
fixed emptyDegree constructor for setting gradeScale

pedrorijo91:


jcarvalho:
This bug was introduced with the fix for ACDM-713.

PR #726, opened 2015-05-04
Further changes to group names to avoid groups sharing the same name ACDM-715 #resolve

gil-l:
Added more info to the execution course part of the Teacher, TeacherResponsibleOfExecutionCourse and Student Group names.
Added info on whether a UnitGroup includes subunits to it's name. Also corrected a bug with the relation type part of the UnitGroup's name.


pedrosan7os:
Squash the commits

gil-l:
Squashed :)

PR #730, opened 2015-05-06
Fix create admin office form for empty slots ACDM-726 #resolves

pedrorijo91:


src/main/webapp/manager/createAcademicOffice.jsp
```diff
@@ -34,18 +34,18 @@
 <fr:edit id="office" name="bean" action="/manageAssociatedObjects.do?method=createAcademicOffice">
     <fr:schema bundle="MANAGER_RESOURCES"
                type="org.fenixedu.academic.ui.struts.action.manager.ManageAssociatedObjects$AssociatedObjectsBean">
-        <fr:slot name="type" key="documents.type">
+        <fr:slot name="type" key="documents.type" validator="pt.ist.fenixWebFramework.renderers.validators.RequiredValidator">
```
jcarvalho: You could just use `required="true"`, it has the same effect as declaring this validator, and is more readable.

PR #732, opened 2015-05-11
Add external id to person payments endpoint. ACDM-729 #resolve

sergiofbsilva:
Add external id to payment events returned by person payments endpoint.

Issue: ACDM-729


jcarvalho:
Be **very** careful when publishing FF's External ID, as there are no real long-term guarantees that those will not change.

pedrosan7os:
Since there is no viable alternative, a short lived id is better than no id. And since this id is primarily used in a short lived compatible manner this is ok for now.

PR #734, opened 2015-05-13
Refactor: Added slots to the RegistrationDataByExecutionYear to support reingression information (and related interfaces)

nadiramin:


nadiramin:
Closes ACM-649

src/main/java/org/fenixedu/academic/domain/student/Registration.java
```diff
@@ -3430,6 +3430,14 @@ public boolean hasMissingPersonalInformationForAcademicService(ExecutionYear exe
...
 
+    @Atomic
+    public void createReingression(ExecutionYear executionYear, LocalDate reingressionDate) {
+        RegistrationDataByExecutionYear dataByYear =
+                RegistrationDataByExecutionYear.getOrCreateRegistrationDataByYear(this, executionYear);
+        dataByYear.createReingression(reingressionDate);
+    }
+
+    @Deprecated
     public boolean isReingression(final ExecutionYear executionYear) {
```
jcarvalho: Given that this PR is targeting a major version, you may safely delete this method (it seems to me that it is unused anyway).

src/main/java/org/fenixedu/academic/domain/student/RegistrationDataByExecutionYear.java
```diff
@@ -67,7 +102,7 @@ public static RegistrationDataByExecutionYear getOrCreateRegistrationDataByYear(
                         .filter(registrationDataByExecutionYear -> registrationDataByExecutionYear.getExecutionYear() == executionYear)
                         .findAny();
 
-        return result.orElse(new RegistrationDataByExecutionYear(registration, executionYear));
+        return result.isPresent() ? result.get() : new RegistrationDataByExecutionYear(registration, executionYear);
```
jcarvalho: You may leave this as is, but a cleaner way of accomplishing this would be to use the `orElseGet` method of the `Optional` class.

src/main/dml/fenixedu-academic.dml
```diff
@@ -2921,6 +2921,8 @@ valueType org.fenixedu.academic.domain.accounting.EventTypes as EventTypes {
     }
 
     class student.RegistrationDataByExecutionYear  {
+		Boolean reingression;
```
pedrosan7os: Seems that the semantic you are looking for here would be better supported with primitive boolean (you are turning nulls into false).

PR #736, opened 2015-05-15
Fix edit degreeCurricularPlan in coordinator menu ACDM-771 #resolve

pedrorijo91:


src/main/webapp/coordinator/degreeSite/viewDescriptionCurricularPlan.jsp
```diff
@@ -40,14 +42,11 @@
...
-		<fr:edit name="degreeCurricularPlan" type="org.fenixedu.academic.domain.DegreeCurricularPlan" schema="degree.curricular.plan.description" nested="true">
-		    <fr:layout>
-	    	    <fr:property name="classes" value="tstyle2 thlight thright thtop mbottom2"/>
-	        	<fr:property name="columnClasses" value="width10em,pbottom1,tdclear tderror1 valigntop"/>
-		    </fr:layout>
-		</fr:edit>
-	
+
+		<textarea bennu-html-editor bennu-localized-string name="degreeCurricularPlanDescription" id="degreeCurricularPlanDescription">
+			${degreeCurricularPlanDescription.json()}
```
jcarvalho: Due to encoding issues, and the fact that inside the JSON can be valid HTML, you must escape the JSON when showing it in the textarea, similar to what is done [here](https://github.com/FenixEdu/fenixedu-cms/blob/v1.0.7/src/main/webapp/WEB-INF/fenixedu-cms/editPost.jsp#L86-L87).

src/main/java/org/fenixedu/academic/ui/struts/action/coordinator/DegreeSiteManagementDispatchAction.java
```diff
@@ -137,16 +144,38 @@ public ActionForward viewDescriptionCurricularPlan(ActionMapping mapping, Action
             return new ActionForward(mapping.getInput());
         }
         request.setAttribute("degreeCurricularPlan", degreeCurricularPlan);
+        request.setAttribute(
+                "degreeCurricularPlanDescription",
+                new LocalizedString().with(Locale.ENGLISH, degreeCurricularPlan.getDescriptionEn()).with(new Locale("pt", "PT"),
```
pedrosan7os: You can use MultiLanguageString.en and MultiLanguageString.pt to access these 'well known' locales.

NunoPinheiro: I would suggest using Locale.ENGLISH and Locale.getDefault() to drop the dependency of "Portuguese"

jcarvalho: And who ensures that the default locale is not English? :)

The actual dependency is in the DegreeCurricularPlan, not here.

PR #751, opened 2015-05-29
Really comparing the states of different enrolments ACDM-689

ricardorcr:


sergiofbsilva:
There should be a reference to the issue in your commit.

PR #765, opened 2015-06-05
Escaped EL expressions using c:out

jcarvalho:
As regular EL expressions do not provide any sort of XML escaping, their single use is prone to XSS attacks.

This PR addresses this issue, but wrapping the relevant expressions in `<c:out>` tags, which properly escape their contents.

The changes in this PR have been tested with the JSPC compiler.

Issue: ACDM-790


pedrosan7os:
Isn't there an issue with this?

jcarvalho:
Due to the severity of the issue solved by this PR, its creation has been defered as close as possible to the next release. The PR has been edited with issue info.

PR #785, opened 2015-06-17
Add methods that return descriptions as LocalizedString

nadiramin:
Relates to https://jira.fenixedu.org/browse/ACDM-651

### DIFF
--- a/src/main/java/org/fenixedu/academic/domain/Degree.java
+++ b/src/main/java/org/fenixedu/academic/domain/Degree.java
@@ -58,7 +58,9 @@ import org.fenixedu.academic.util.Bundle;
 import org.fenixedu.academic.util.MultiLanguageString;
 import org.fenixedu.bennu.core.domain.Bennu;
 import org.fenixedu.bennu.core.i18n.BundleUtil;
+import org.fenixedu.bennu.core.util.CoreConfiguration;
 import org.fenixedu.commons.i18n.I18N;
+import org.fenixedu.commons.i18n.LocalizedString;
 import org.fenixedu.spaces.domain.Space;
 import org.joda.time.DateTime;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/fenixedu/academic/domain/Degree.java
```diff
@@ -615,6 +617,20 @@ final public MultiLanguageString getNameI18N() {
     final public MultiLanguageString getNameI18N(ExecutionYear executionYear) {
         return getNameFor(executionYear);
     }
+    
+    public LocalizedString getPresentationNameI18N() {
+        return getPresentationNameI18N(ExecutionYear.readCurrentExecutionYear());
+    }
+    
+    public LocalizedString getPresentationNameI18N(final ExecutionYear executionYear) {
```
pedrosan7os: This is mostly a reconstruction of a localized string from broken bits of what should be a localized string since the beginning. Make the getPresentationNameI18N the source method from all getPresentationName* methods, and implement it like so:

```java
    public LocalizedString getPresentationNameI18N(final ExecutionYear executionYear) {
        LocalizedString in = BundleUtil.getLocalizedString(Bundle.APPLICATION, "label.in");
        LocalizedString degreeType = getDegreeType().getName();
        Builder builder = degreeType.builder();
        degreeType.getLocales().forEach(l -> builder.append(in.getContent(l), " "));
        builder.append(getNameFor(executionYear).toLocalizedString(), " ");
        return builder.build();
    }
```

All other just invoke this, with variations of ```ExecutionYear.readCurrentExecutionYear()``` as a parameter or ```.getContent()``` invoked after.

src/main/java/org/fenixedu/academic/domain/degreeStructure/CycleType.java
```diff
@@ -157,4 +159,14 @@ public boolean hasPrevious() {
         return getPrevious() != null;
     }
 
+    public LocalizedString getDescriptionI18N() {
```
pedrosan7os: Again, and for the same reasons, getDescriptionI18N should be the source:

```java
    public LocalizedString getDescriptionI18N() {
        return BundleUtil.getLocalizedString(Bundle.ENUMERATION, getQualifiedName());
    }
```

and ```getDescription(Locale)``` should just invoke this with ```.getContent(locale)``` after

src/main/java/org/fenixedu/academic/domain/person/IDDocumentType.java
```diff
@@ -65,4 +67,15 @@ public String getLocalizedName() {
     public String getLocalizedName(final Locale locale) {
         return BundleUtil.getString(Bundle.ENUMERATION, locale, name());
     }
+    
+    public LocalizedString getLocalizedNameI18N() {
```
pedrosan7os: Same here:

```java
    public LocalizedString getLocalizedNameI18N() {
        return BundleUtil.getLocalizedString(Bundle.ENUMERATION, name());
    }
```

pedrosan7os:
Fixed in https://github.com/FenixEdu/fenixedu-academic/commit/91bb361f7be2c279ef33e8a43dcb1bda2b4dc939

PR #789, opened 2015-06-22
Updated the candidacy process. ACDM-706 #resolve

cfscosta:
Added: candidacy processes can be created with overlapping periods.
Added: Its is now possible to move candidates between candidacy processes.


pedrosan7os:
Now properly integrated with the existing renderers UI, thank you very much :clap:

PR #791, opened 2015-06-23
Deletion of unused schema file

ricardorcr:


pedrosan7os:
From the academic point of view this is just a deletion of unused resources. No need to refer a non academic issue. The commit and pull request descriptions should also follow the same logic.

PR #792, opened 2015-06-23
Removed the exposure of OID's in the reports

ricardorcr:


pedrosan7os:
From the academic standpoint this is deprecated, to be removed in the next major.

PR #796, opened 2015-06-26
Fixed space management issues. ACDM-826 #resolve

cfscosta:
When a person is authorized to manage a space, the required interfaces
and authorizations are set/added automatically.


jcarvalho:
- In what way does this change help fix the issue?
- Shouldn't this be in the Spaces module?

PR #811, opened 2015-07-09
Checks for null input stream on project submission file ACDM-888 #resolve

diutsu:


jcarvalho:
How will this PR help solve the issue?

First, the only reason that the `InputStream` is null is if the user manually canceled the upload, and even then, this changeset wouldn't prevent a NPE.

diutsu:
The ByteStreams.toByteArray can throw a NPE if 'is' is null.
That was what happened in the external issue associated.

jcarvalho:
I know, you probably missed a `return` or an `else` statement in that `if`, as if the stream is `null`, you're still invoking the `toByteArrayMethod`.

diutsu:
Should be fixed now, thanks.

pedrosan7os:
This should refer to an issue of the same project. FIST-95 is not.

PR #820, opened 2015-07-20
It is now possible to select the semester in which the student wants to enroll ACDM-838 #resolve

ricardorcr:
- this was applied to classes and courses enrolment
- if there is only one active enrolment period the semester it is automatically choosen

### DIFF
--- a/src/main/java/org/fenixedu/academic/domain/DegreeCurricularPlan.java
+++ b/src/main/java/org/fenixedu/academic/domain/DegreeCurricularPlan.java
@@ -29,6 +29,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
@@ -641,6 +642,13 @@ public class DegreeCurricularPlan extends DegreeCurricularPlan_Base {
         return null;
     }
 
+    public Optional<EnrolmentPeriod> getActiveCurricularCourseEnrolmentPeriod(ExecutionSemester executionSemester) {
+        return getEnrolmentPeriodsSet()
+                .stream()
+                .filter(ep -> ep.getExecutionPeriod() == executionSemester && ep instanceof EnrolmentPeriodInCurricularCourses
+                        && ep.isValid()).findAny();
+    }
+
     public boolean hasActualEnrolmentPeriodInCurricularCourses() {
         return getActualEnrolmentPeriod() != null;
     }
@@ -667,6 +675,7 @@ public class DegreeCurricularPlan extends DegreeCurricularPlan_Base {
         return false;
     }
 
+    //TODO remove in next major it's no used and there can be more than one active, if don't specify the semester
     public EnrolmentPeriodInCurricularCoursesSpecialSeason getActualEnrolmentPeriodInCurricularCoursesSpecialSeason() {
         for (final EnrolmentPeriod enrolmentPeriod : this.getEnrolmentPeriodsSet()) {
             if (enrolmentPeriod instanceof EnrolmentPeriodInCurricularCoursesSpecialSeason && enrolmentPeriod.isValid()) {
@@ -674,6 +683,13 @@ public class DegreeCurricularPlan extends DegreeCurricularPlan_Base {
             }
         }
         return null;
+    }
+
+    public Optional<EnrolmentPeriod> getActiveEnrolmentPeriodInCurricularCoursesSpecialSeason(ExecutionSemester executionSemester) {
+        return getEnrolmentPeriodsSet()
+                .stream()
+                .filter(ep -> ep.getExecutionPeriod() == executionSemester
+                        && ep instanceof EnrolmentPeriodInCurricularCoursesSpecialSeason && ep.isValid()).findAny();
     }
 
     public boolean hasOpenEnrolmentPeriodInCurricularCoursesSpecialSeason(final ExecutionSemester executionSemester) {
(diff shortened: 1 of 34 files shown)


src/main/java/org/fenixedu/academic/domain/DegreeCurricularPlan.java
```diff
@@ -773,6 +773,15 @@ public EnrolmentPeriodInClasses getCurrentClassesEnrollmentPeriod() {
         return null;
     }
 
+    public EnrolmentPeriodInClasses getClassesEnrollmentPeriod(ExecutionSemester executionSemester) {
```
jcarvalho: This would be a great candidate for an `Optional`.

src/main/java/org/fenixedu/academic/domain/EnrolmentPeriod.java
```diff
@@ -83,6 +83,10 @@ protected void init(final DegreeCurricularPlan degreeCurricularPlan, final Execu
         setEndDateDateTime(endDate);
     }
 
+    public abstract boolean isForCurricularCourses();
```
jcarvalho: Adding this is an API break. You should provide default implementations for these methods.

jcarvalho: I suggest having them both return `false`, and only override to `true` where needed.

src/main/java/org/fenixedu/academic/servlet/taglib/sop/v3/RenderTimeTableTag.java
```diff
@@ -74,6 +74,8 @@
 
     private String executionCourseID = "";
 
+    private String executionSemesterOID = "";
```
jcarvalho: I think we should avoid using the `OID` suffix for properties.

src/main/webapp/student/enrollment/chooseRegistration.jsp
```diff
@@ -38,19 +38,20 @@
...
     	</p>
     </logic:empty>
+    <bean:define id="executionSemesterOID" name="executionSemester" property="externalId"/>
 	<logic:notEmpty name="registrationsToEnrol">
 		<fr:view name="registrationsToEnrol" schema="student.registrationDetail.short" >
 			<fr:layout name="tabular">
 				<fr:property name="sortBy" value="startDate=desc"/>			
 				<fr:property name="classes" value="tstyle4 thlight mtop025 indent1"/>
-				<fr:property name="linkFormat(enrol)" value="/studentEnrollmentManagement.do?method=chooseRegistration&registrationId=${externalId}" />
+				<fr:property name="linkFormat(enrol)" value="<%= "/studentEnrollmentManagement.do?method=chooseRegistration&registrationId=${externalId}" + "&executionSemesterOID="executionSemesterOID %>"/>
```
jcarvalho: Why are you mixing EL expressions with the `<%=` notation?

src/main/java/org/fenixedu/academic/domain/CandidacyPeriodInDegreeCurricularPlan.java
```diff
@@ -48,5 +48,4 @@ private void checkRuleToCreate(final DegreeCurricularPlan degreeCurricularPlan,
     public ExecutionYear getExecutionYear() {
         return getExecutionPeriod().getExecutionYear();
     }
-
```
pedrosan7os: white space

src/main/java/org/fenixedu/academic/service/filter/enrollment/ClassEnrollmentAuthorizationFilter.java
```diff
@@ -108,19 +146,21 @@ private FenixServiceException verify(StudentCurricularPlan studentCurricularPlan
...
     }
 
-    private FenixServiceException verify(DegreeCurricularPlan degreeCurricularPlan) {
-        final EnrolmentPeriodInClasses enrolmentPeriodInClasses = degreeCurricularPlan.getCurrentClassesEnrollmentPeriod();
-        if (enrolmentPeriodInClasses == null || enrolmentPeriodInClasses.getStartDateDateTime() == null
-                || enrolmentPeriodInClasses.getEndDateDateTime() == null) {
+    private FenixServiceException verify(DegreeCurricularPlan degreeCurricularPlan, ExecutionSemester executionSemester) {
+        final Optional<EnrolmentPeriod> enrolmentPeriodInClasses =
+                degreeCurricularPlan.getClassesEnrollmentPeriod(executionSemester);
+        if (!enrolmentPeriodInClasses.isPresent() || enrolmentPeriodInClasses.get().getStartDateDateTime() == null
```
pedrosan7os: Optional has a filter method that can replace your ifPresent + get pattern

src/main/java/org/fenixedu/academic/service/filter/enrollment/ClassEnrollmentAuthorizationFilter.java
```diff
@@ -108,19 +146,21 @@ private FenixServiceException verify(StudentCurricularPlan studentCurricularPlan
...
+                || enrolmentPeriodInClasses.get().getEndDateDateTime() == null) {
             return new CurrentClassesEnrolmentPeriodUndefinedForDegreeCurricularPlan();
         }
 
-        if (!enrolmentPeriodInClasses.isValid()) {
+        if (!enrolmentPeriodInClasses.get().isValid()) {
```
pedrosan7os: Is this de-reference checked?

ricardorcr: Yes, if it reaches this condition it is because it has a value

src/main/java/org/fenixedu/academic/service/filter/enrollment/ClassEnrollmentAuthorizationFilter.java
ricardorcr: It is possible to use some of that but in this case there are 3 different possible outcomes and it would be somewhat confusing doing it through that way in the entire method

PR #824, opened 2015-07-21
CurricularCourse no longer assumes default values for Credits and Weight

jcarvalho:
* Issues: ACDM-752, ACDM-751

### DIFF
--- a/src/main/java/org/fenixedu/academic/domain/CurricularCourse.java
+++ b/src/main/java/org/fenixedu/academic/domain/CurricularCourse.java
@@ -68,10 +68,6 @@ import org.joda.time.DateTime;
 
 public class CurricularCourse extends CurricularCourse_Base {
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/fenixedu/academic/domain/CurricularCourse.java
```diff
@@ -1067,8 +1065,7 @@ final public Double getWeight() {
             return getCredits();
         }
 
-        final Double weigth = super.getWeigth();
-        return (weigth == null || weigth == 0.0) ? WEIGHT_FOR_PRE_BOLONHA : weigth;
```
sergiofbsilva: shouldn't the constants be removed from the source code?

jcarvalho: You're right. Fixed.

PR #835, opened 2015-07-28
Add possibility to cancel Enrolments.

nadiramin:
Closes ACDM-858


src/main/java/org/fenixedu/academic/ui/struts/action/administrativeOffice/studentEnrolment/StudentEnrolmentsDA.java
```diff
@@ -102,4 +104,44 @@ public ActionForward backViewRegistration(ActionMapping mapping, ActionForm form
         return mapping.findForward("visualizeRegistration");
     }
 
+    public ActionForward annulEnrolment(final ActionMapping mapping, final ActionForm form, final HttpServletRequest request,
+            final HttpServletResponse response) {
+
+        try {
+            atomic(() ->
```
jcarvalho: I think it would be better to reduce the amount of code inside these atomic blocks. I would extract the `enrolment` variable outside of it, and the just use `atomic(enrolment::annul)`.

src/main/webapp/academicAdminOffice/chooseStudentEnrolmentExecutionPeriod.jsp
```diff
@@ -23,6 +23,8 @@
 <%@ taglib uri="http://struts.apache.org/tags-logic" prefix="logic"%>
 <%@ taglib uri="http://fenix-ashes.ist.utl.pt/fenix-renderers" prefix="fr" %>
 <%@ taglib uri="http://fenix-ashes.ist.utl.pt/taglib/academic" prefix="academic" %>
+<%@ page isELIgnored="true"%>
```
jcarvalho: You could avoid this by just escaping the new `${}` entries you are creating :smile:

pedrosan7os:
Fixed in https://github.com/FenixEdu/fenixedu-academic/commit/104d459a5305acab2d58f76f9deacf1365435ce7

PR #837, opened 2015-07-29
Fixed message sending problem when body is null ACDM-781 #resolve

cfscosta:
When the message body is null the system will skip the message and will
show the number of recipients as 0.

### DIFF
--- a/src/main/java/org/fenixedu/academic/domain/util/email/Message.java
+++ b/src/main/java/org/fenixedu/academic/domain/util/email/Message.java
@@ -34,6 +34,8 @@ import org.joda.time.DateTime;
 import pt.ist.fenixframework.Atomic;
 import pt.ist.fenixframework.Atomic.TxMode;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/fenixedu/academic/domain/util/email/Message.java
```diff
@@ -264,6 +264,15 @@ public String getRecipientsGroupMembersInText() {
...
     public int dispatch() {
+        boolean subject = (getSubject() == null || getSubject().isEmpty());
+        if (subject) {
+            return 0;
+        }
+        boolean body = (getBody() == null || getBody().isEmpty());
+        boolean htmlBody = (getHtmlBody() == null || getHtmlBody().isEmpty());
+        if (body || htmlBody) {
+            return 0;
+        }
```
sergiofbsilva: why not a single if with `Strings.isNullOrEmpty` ?

PR #840, opened 2015-07-30
New interface for registration transfer between students. ACDM-856 #resolve

pedrosan7os:
- The interface is accessible to CREATE_REGISTRATIONS academic authorization
 - All academic relevant information is transfered from a source to a target student
 - Nothing is done to the actual accounts
 - Several setters, mainly in the payments domain, that were protected by an exception throw, had to be opened


src/main/java/org/fenixedu/academic/domain/accounting/paymentCodes/AccountingEventPaymentCode.java
```diff
@@ -138,15 +138,6 @@ public void delete() {
     }
 
     @Override
-    public void setPerson(Person student) {
-        throw new DomainException("error.org.fenixedu.academic.domain.accounting.PaymentCode.cannot.modify.person");
-    }
-
-    protected void _setPerson(Person person) {
```
jcarvalho: Mark this method as `@Deprecated` instead.

pedrosan7os: Can you motivate your advice?

PR #841, opened 2015-07-31
Checking access control before selection menu item in StrutsFunctiolityController. ACDM-899 #resolve

pedrosan7os:


src/main/java/org/fenixedu/academic/ui/spring/StrutsFunctionalityController.java
```diff
@@ -34,7 +35,10 @@
     @ModelAttribute
     private void selectStrutsFunctionality(HttpServletRequest request) {
         Functionality functionality = RenderersAnnotationProcessor.getFunctionalityForType(getFunctionalityType());
-        BennuPortalDispatcher.selectFunctionality(request,
-                MenuFunctionality.findFunctionality(functionality.getProvider(), functionality.getKey()));
+        MenuFunctionality menuItem = MenuFunctionality.findFunctionality(functionality.getProvider(), functionality.getKey());
+        if (!menuItem.isAvailableForCurrentUser()) {
```
jcarvalho: You should add a `null` check here, as the selected functionality may not be installed, yielding a `null` Menu Item.

pedrosan7os: Included the null check with the semantic of unauthorized if null, not having the functionality installed in the menu means its unavailable

jcarvalho:
:+1:

PR #845, opened 2015-08-05
Minor fix: removing hardcoded "academica@tecnico.pt". Added new domain configuration to replace these cases

nadiramin:
Closes ACDM-903


jcarvalho:
You should add a fallback if the new slot is not filled (like overridding the getter method with the previous default value), and leave a note on the new slot for it to be marked as required in the next major.

PR #848, opened 2015-08-14
Added caching headers to the legacy photo actions

jcarvalho:
* Issue: ACDM-912

### DIFF
--- a/src/main/webapp/person/visualizePersonalInfo.jsp
+++ b/src/main/webapp/person/visualizePersonalInfo.jsp
@@ -64,7 +64,7 @@
 			<tr>
 				<td align="center">
 					<div class="items-container" data-toggle="tooltip" data-placement="right" title="${fr:message('resources.ApplicationResources', 'link.operator.submitPhoto')}">
-						<img src="${fr:checksum('/person/retrievePersonalPhoto.do?method=retrieveOwnPhoto')}" width="100" height="100" align="middle" style="border: 2px #eee solid"/>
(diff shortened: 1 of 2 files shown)


src/main/webapp/person/visualizePersonalInfo.jsp
```diff
@@ -64,7 +64,7 @@
 			<tr>
 				<td align="center">
 					<div class="items-container" data-toggle="tooltip" data-placement="right" title="${fr:message('resources.ApplicationResources', 'link.operator.submitPhoto')}">
-						<img src="${fr:checksum('/person/retrievePersonalPhoto.do?method=retrieveOwnPhoto')}" width="100" height="100" align="middle" style="border: 2px #eee solid"/>
+						<img src="${fr:checksum('/person/retrievePersonalPhoto.do?method=retrieveOwnPhoto&'.concat(person.personalPhotoEvenIfPending.externalId))}" width="100" height="100" align="middle" style="border: 2px #eee solid"/>
```
pedrosan7os: If you need to pass the id then there is no point in having a method retrieveOwnPhoto. Probably there is a method by id that suits your need, and then delete the retrieveOwnPhoto, one less method to take care of.

jcarvalho: In this invocation, the id is just included for Cache Busting, it is not actually used.

jcarvalho:
In second thought, I think I can come up with a better solution. Closing for now.

PR #849, opened 2015-08-14
Allow generation of student numbers above 100000. ACDM-893 #resolve

sanavarali:
Closes ACDM-911


pedrosan7os:
The correct issue would be ACDM-893

PR #852, opened 2015-08-18
added an empty option to enable degree selection, even if there is only one degree available

sanavarali:
closes #ACDM-919


jcarvalho:
Did you check that usages of this bean can handle the property being `null`?

sanavarali:
The property won't be `null`. The `RequiredValidator` for executionDegree slot is still there.

PR #863, opened 2015-08-27
Minor Fix: Allow support for usernames with dots, in PhotographController

nadiramin:
Closes ACDM-926


jcarvalho:
Funny part about this is that only the last dot is 'eaten', all the others were being maintained.

NunoPinheiro:
What I understood from the research I did at the time, Spring considers the contents after the last dot to be the extension of the file so "my.file.extension" would be recieved as "my.file" and the extension would be dropped.

PR #890, opened 2015-09-29
Bug fix displaying error and warning messages

nadiramin:
Closes ACDM-913


jcarvalho:
This does not seem like the proper solution to this problem.

You are trying to address some non-localized warning messages by removing the pattern added by Struts when a key is not found, right?

The best solution for this would be to either properly treat the errors as literals in Java, or don't use literals at all.

PR #891, opened 2015-09-29
Delay the verification of diploma or registry certificate in process action instead in degree finalization certificate request, in order to register and receive the emolument payment

nadiramin:
Closes ACDM-922


pedrosan7os:
This is a disruptive change on the current semantics of the process. A direct change like this would have to follow a more thorough discussion.

nadiramin:
This behavior may change by institution, university or country, I would suggest changing this to an extension point / configurable behavior to allow each instance to decide what makes sense.

New issue for this extension point?

pedrosan7os:
Agreed

PR #897, opened 2015-10-05
[JIRA] (ACDM-969) Improvement: Possibility to plug new enrolment managers

nadiramin:
Closes ACDM-969.

### DIFF
--- a/src/main/java/org/fenixedu/academic/domain/studentCurriculum/StudentCurricularPlanEnrolment.java
+++ b/src/main/java/org/fenixedu/academic/domain/studentCurriculum/StudentCurricularPlanEnrolment.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import org.fenixedu.academic.domain.DegreeCurricularPlan;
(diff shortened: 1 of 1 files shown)


src/main/java/org/fenixedu/academic/domain/studentCurriculum/StudentCurricularPlanEnrolment.java
```diff
@@ -380,31 +381,45 @@ protected boolean isResponsiblePersonCoordinator() {
     // -------------------
 
     static public StudentCurricularPlanEnrolment createManager(final EnrolmentContext enrolmentContext) {
+        return ENROLMENT_MANAGER_FACTORY.get().createManager(enrolmentContext);
+    }
 
-        if (enrolmentContext.isNormal()) {
+    static public Supplier<EnrolmentManagerFactory> ENROLMENT_MANAGER_FACTORY = () -> new EnrolmentManagerFactory() {
```
jcarvalho: Don't publicly expose an internal field, keep it `private` and expose a setter (which checks that the new factory is not `null`).

legidio:
Will do.

As a side note, maybe the same suggested checks should be added in this extension point:

https://github.com/FenixEdu/fenixedu-academic/blob/master/src/main/java/org/fenixedu/academic/domain/student/curriculum/Curriculum.java#L162

pedrosan7os:
Accepted suggestion: a49e8fe4d52fe11bf51062af5b3b378159ee8afe

legidio:
Done.

Applied same rationale, with the adition of also testing for the Supplier's result, obtained from the get() method.

PR #901, opened 2015-10-09
Improvement, Merge Execution Course: allow to distinguish ExecutionCourses of the same CompetenceCourse under the same Degree but from different DegreeCurricularPlan

nadiramin:


src/main/java/org/fenixedu/academic/ui/renderers/providers/DestinationExecutionCoursesProvider.java
```diff
@@ -18,58 +18,71 @@
...
+            public Object convert(final Class type, final Object value) {
+                final ExecutionCourseBean result = new ExecutionCourseBean();
+                result.setDestinationExecutionCourse((ExecutionCourse) new DomainObjectKeyConverter().convert(type, value));
+                return result;
+            }
 
-        Collections.sort(destinationExecutionCourses, DomainObjectUtil.COMPARATOR_BY_ID);
+            @Override
+            public String deserialize(final Object object) {
+                String result = StringUtils.EMPTY;
```
jcarvalho: Just use `""` instead. It's just as clear as the constant for anyone reading this, and avoids another dependency. (The same goes for the other usages in this PR).

src/main/java/org/fenixedu/academic/ui/struts/action/academicAdministration/executionCourseManagement/ExecutionCourseBean.java
```diff
@@ -122,7 +149,62 @@ private DegreeCurricularPlan getDegreeCurricularPlan() {
         return getExecutionDegree().getDegreeCurricularPlan();
     }
 
-    public ExecutionCourseBean() {
+    public String getSourcePresentationName() {
+        String result = StringUtils.EMPTY;
+
+        if (getSourceExecutionCourse() != null) {
+            result += getSourceExecutionCourse().getNameI18N().getContent();
```
jcarvalho: This should use a `StringBuilder`.

src/main/java/org/fenixedu/academic/ui/struts/action/academicAdministration/executionCourseManagement/ExecutionCourseBean.java
```diff
@@ -122,7 +149,62 @@ private DegreeCurricularPlan getDegreeCurricularPlan() {
...
+
+        return result;
+    }
+
+    static private String getDegreeCurricularPlansPresentationString(final Set<DegreeCurricularPlan> input) {
+        String result = " [ ";
+
+        for (Iterator<DegreeCurricularPlan> iterator = input.iterator(); iterator.hasNext();) {
+            final DegreeCurricularPlan iter = iterator.next();
+            result += iter.getName();
```
jcarvalho: This should use a `StringBuilder` (the same goes for all the other usages of `+=` on Strings for this PR).

legidio:
Done.

My mistake. I knew the StringUtils' "unofficial deprecation" from code-inspecting the many refactorings over the last 2 years. By the way, is there a code convention guide for the project where someone from the community can get to know this and other code best practises?

As for the StringBuilder, I (wrongly?) assumed the compiler's code optimization that internaly replaces string concatenation with StringBuilder.

I'm glad that no domain logic bug was found and that an old UI is greatly improved, for the benefit of the final user.

jcarvalho:
> As for the StringBuilder, I (wrongly?) assumed the compiler's code optimization that internaly replaces string concatenation with StringBuilder.

Actually, the compiler will create a new instance of `StringBuilder` every time you do a `+=`, so using it in succession in the same method would actually create multiple instances of `StringBuilder`.

PR #909, opened 2015-10-19
Enable skipping physical address validation depending on configuration. Minor refactor: logic to check if contact must be validated is now inside each type

sanavarali:
resolves ACDM-956


jcarvalho:
Integrated in #950.

PR #913, opened 2015-10-19
Fixed bug in CompetenceCourse comparator by name

sanavarali:
resolves ACDM-906


pedrosan7os:
The solution does not conform with what is being discussed in the issue

PR #934, opened 2016-01-12
 ACDM-1020 #resolve Fix NPE when enrollment end time was close to midnight

diutsu:
Due to conversions between java.util.Date java.util.Calendar and
org.joda.time.DateTime an extra 13 seconds where being added.
This made the interval computation fail when the end time was close to the
midnight

### DIFF
--- a/src/main/java/org/fenixedu/academic/dto/student/StudentPortalBean.java
+++ b/src/main/java/org/fenixedu/academic/dto/student/StudentPortalBean.java
@@ -20,7 +20,6 @@ package org.fenixedu.academic.dto.student;
 
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.Calendar;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/org/fenixedu/academic/dto/student/StudentPortalBean.java
```diff
@@ -241,29 +239,14 @@ public void setRealization(Grouping grouping) {
...
-                final DateTime beginDateTime = toDateTime(beginDay, beginTime);
-                final DateTime endDateTime = toDateTime(endDay, endTime);
+                DateTime beginDateTime = writtenEvaluation.getEnrolmentPeriodStart();
+                DateTime endDateTime = writtenEvaluation.getEnrolmentPeriodEnd();
 
-                Interval interval = new Interval(beginDateTime, endDateTime);
+                this.enrolmentPast = endDateTime == null ? false : endDateTime.isBeforeNow();
 
-                this.enrolmentElapsing = interval.containsNow();
+                this.enrolmentElapsing = new Interval(beginDateTime, endDateTime).containsNow();
```
jcarvalho: Won't the Interval constructor throw a NPE if its arguments are `null`?

diutsu:
apparently no, 
http://joda-time.sourceforge.net/apidocs/org/joda/time/Interval.html#Interval(org.joda.time.ReadableInstant,%20org.joda.time.ReadableInstant)
>  null means now

at most we get could get an IllegalArgumentException, if the end is before the start

PR #947, opened 2016-03-03
Removed different behaviour retrieving Competence name ACDM-1045 #resolve

ricardorcr:
The way of getting a Competence Course name is now uniform for all


jcarvalho:
Won't this change have a massive impact on other functionalities that use the `CompetenceCourseInformation`?

This seems a bit too sensitive to be done in a revision...

sergiofbsilva:
CurricularStage.OLD is only used by the old domain so this is has little impact. The method isBolonha should be removed in a major version. The impact of this change is minimal. The side effect of changing the CurricularStage of to a different value in all OLD competences is unknown.

pedrosan7os:
Agree, however it should at least be an improvement, not a simple bug fix.

PR #949, opened 2016-03-15
Dealing with senders when merging execution courses ACDM-1027 #resolve

ricardorcr:


src/main/java/org/fenixedu/academic/service/services/manager/MergeExecutionCourses.java
```diff
@@ -424,6 +426,14 @@ private static void copyPersistentGroups(ExecutionCourse executionCourseFrom, Ex
         }
     }
 
+    private static void copySenderMessages(ExecutionCourse executionCourseFrom, ExecutionCourse executionCourseTo) {
+        if (executionCourseFrom.getSender() != null) {
+            ExecutionCourseSender courseSenderTo = ExecutionCourseSender.newInstance(executionCourseTo);
+            courseSenderTo.getMessagesSet().addAll(executionCourseFrom.getSender().getMessagesSet());
+            executionCourseFrom.getSender().delete();
```
jcarvalho: What about the recipients?

ricardorcr: The recipients that exist in an execution course are computed based on the execution course itself so there is no need to copy the recipients, it is only necessary to copy the messages that were sent.

jcarvalho: From the code in the Recipient's `delete` method, any message associated with it, will be deleted.

As this does not dissociate those messages from the recipients, they will be deleted.

PR #950, opened 2016-04-06
New configuration property to control physical address validation

pedrosan7os:
ACDM-956 #resolve


NunoPinheiro:
There's already another pull request fixing this issue:
https://github.com/FenixEdu/fenixedu-academic/pull/909

I would advice refactoring the code a bit as presented in that pull request since the logic is becoming kind of fuzzy.

pedrosan7os:
The mentioned pull request had a number of issues. The more serious being its scope: it changed more than the related issue would justify. This solution adds the exact same feature with a fraction of the change set.

But if you want to discuss the extra changes, namely on `isToBeValidated` method, I would argue that spreading a logic operation through a class hierarchy with lots of unnecessary overrides is not doing any good to the codebase. The core issue is that the pattern of having lots of subclasses with overrides is in fact an anti-pattern, that no one wishes to perpetuate.

PR #954, opened 2016-04-18
Manage accounts interface problems ACDM-642 #resolve

fernandojps:
ACDM-642 was being caused by a null nationality, when trying to get `nationality.localizedName.content`. In order to fix this, the results table had to be re-written without renderers.


fernandojps:
I found another bug, this is not ready to be merged yet.

PR #955, opened 2016-04-18
Manage accounts interface problems ACDM-642 #resolve

fernandojps:
ACDM-642 was being caused by a null nationality, when trying to get `nationality.localizedName.content`. In order to fix this, the results table had to be re-written without renderers.

### DIFF
--- a/src/main/webapp/accounts/createPerson.jsp
+++ b/src/main/webapp/accounts/createPerson.jsp
@@ -23,6 +23,7 @@
 <%@ taglib uri="http://struts.apache.org/tags-bean" prefix="bean"%>
 <%@ taglib uri="http://struts.apache.org/tags-logic" prefix="logic"%>
 <%@ taglib uri="http://fenix-ashes.ist.utl.pt/fenix-renderers" prefix="fr" %>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c"%>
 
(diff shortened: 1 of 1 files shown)


src/main/webapp/accounts/createPerson.jsp
```diff
@@ -54,24 +54,30 @@
...
+						<th><label><bean:message bundle="MANAGER_RESOURCES" key="label.documentIdNumber"/></label></th>
+						<th><label><bean:message bundle="MANAGER_RESOURCES" key="label.email"/></label></th>
+						<th><label><bean:message bundle="MANAGER_RESOURCES" key="label.dateOfBirth"/></label></th>
+						<th><label><bean:message bundle="MANAGER_RESOURCES" key="label.nationality"/></label></th>
+					</tr>
+				</thead>
+				<tbody>
+					<logic:iterate name="resultPersons" id="resultPerson">
+						<tr>
+							<td><a href='manageAccounts.do?method=viewPerson&amp;personId=<bean:write name="resultPerson" property="externalId"/>'><bean:write name="resultPerson" property="profile.fullName"/>(<bean:write name="resultPerson" property="username"/>)</a></td>
```
jcarvalho: Avoid using relative links, as they can cause some issues with the Portal and Struts's modules. Look into the `fr:checksum` function for that.

PR #957, opened 2016-04-27
Manage Account Email Search. ACDM-1046 #resolve

pasadinhas:
Filter inactive & invalid email addresses, return all matches

### DIFF
--- a/src/main/java/org/fenixedu/academic/domain/Person.java
+++ b/src/main/java/org/fenixedu/academic/domain/Person.java
@@ -1101,15 +1101,6 @@ public class Person extends Person_Base {
         return emailAddress != null && emailAddress.getParty().isPerson() ? (Person) emailAddress.getParty() : null;
     }
 
-    public static Stream<Person> readPeopleByEmailAddress(final String email) {
-        return EmailAddress.findAllActiveAndValid(email)
-                .filter(Objects::nonNull)
-                .map(emailAddress -> emailAddress.getParty())
-                .filter(Objects::nonNull)
-                .filter(party -> party.isPerson())
-                .map(party -> (Person) party);
-    }
-
     public boolean hasEmailAddress(final String email) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/fenixedu/academic/domain/Person.java
```diff
@@ -1101,6 +1101,15 @@ public static Person readPersonByEmailAddress(final String email) {
         return emailAddress != null && emailAddress.getParty().isPerson() ? (Person) emailAddress.getParty() : null;
     }
 
+    public static Stream<Person> readPeopleByEmailAddress(final String email) {
```
jcarvalho: I think the `Person` class is bloated enough as it is, I see no need to add a method that is basically a wrapper, and that is only invoked in a single place.

src/main/java/org/fenixedu/academic/domain/Person.java
```diff
@@ -1101,6 +1101,15 @@ public static Person readPersonByEmailAddress(final String email) {
...
+        return EmailAddress.findAllActiveAndValid(email)
+                .filter(Objects::nonNull)
+                .map(emailAddress -> emailAddress.getParty())
```
jcarvalho: You can just do `.map(EmailAddress::getParty)`.

src/main/java/org/fenixedu/academic/domain/Person.java
```diff
@@ -1101,6 +1101,15 @@ public static Person readPersonByEmailAddress(final String email) {
...
+                .filter(Objects::nonNull)
+                .filter(party -> party.isPerson())
+                .map(party -> (Person) party);
```
jcarvalho: Here you can do `.map(Person.class::cast)` instead.

pasadinhas:
I fixed those issues.

PR #975, opened 2016-06-20
fix student curricular plan selection ACDM-798 #resolve

luisfsantos:
A curricular plan other than the default can now be selected
ACDM-798 #resolve


Luis-Cruz:
Merged manually.

PR #979, opened 2016-06-24
Fix Edit Curricular Course and Course Search interfaces ACDM-1084 #resolve

gil-l:
List all existing contexts in edit curricular course page
Add end semester to context search results
Order search and curricular course context lists
Fix edit curricular course page's context actions redirection
Fix Exception on trying to add new context with no selected course group


Luis-Cruz:
Merged manually.

PR #981, opened 2016-06-29
Change conclusion year calculation in RegistrationConlusionBean ACDM-1086 #resolve

gil-l:
Calculate RegistrationConlusionBean's date based on curriculum group instead of registration


Luis-Cruz:
Merged manually.

PR #982, opened 2016-06-30
Correct RaidesSpecializationReportFile's Degree Type treatment ACDM-1088 #resolve

gil-l:
Delete RaidesSpecializationReportFile degree type getter in favor of the base getter
Override degree type setter to throw IllegalArgumentException if given a non specialization degree type


Luis-Cruz:
Merged manually.

PR #984, opened 2016-07-02
Sort projects also by "end date" and "name" ACDM-829 #resolve

luismrsilva:
This should result on an order that makes more sense when "begin date" is the same.


Luis-Cruz:
Merged manually.

PR #989, opened 2016-07-05
Fix NPE in Registration.getCycleType ACDM-1093 #resolve

gil-l:
Add no cycle curriculum group check to Registration.getCycleType


Luis-Cruz:
Merged manually.

PR #991, opened 2016-07-06
Fix special season enrolment deletion ACDM-1095 #resolve

ricardorcr:
when a trying to disenroll after a debt is generated it was not possible


Luis-Cruz:
Merged manually.

PR #992, opened 2016-07-07
Add quotes to response header in exportToExcel ACDM-1071 #resolve

luisfsantos:
As the filename is not formated it can contain commas. This filename is used to
create the response header as such it is now in double quotes as commas can generate issues.
Resolves #ACDM-1071


Luis-Cruz:
Merged manually.

PR #994, opened 2016-07-08
Allow configuration of admission documents ACDM-1068 #resolve

ricardorcr:


src/main/dml/fenixedu-academic.dml
```diff
@@ -4635,6 +4635,11 @@ public class log.StudentRegistrationTransferLog {
     protected String who;
 }
 
+
+class candidacy.FirstTimeDocumentsConfiguration {
+	JsonElement configurationProperties;
```
jcarvalho: Be careful here, as you should **NEVER** expose `JsonElement`s directly, as they are not immutable and **will** cause issues.

src/main/java/org/fenixedu/academic/domain/candidacy/FirstTimeDocumentsConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Atomic(mode = TxMode.WRITE)
+    private static void initialize() {
+        if (Bennu.getInstance().getFirstTimeDocumentsConfiguration() == null) {
+            new FirstTimeDocumentsConfiguration();
+        }
+    }
+
+    public boolean isToExclude(String property) {
+        JsonElement jsonElement = getInstance().getConfigurationProperties().getAsJsonObject().get(property);
```
sergiofbsilva: this is an instance method, should not use getInstance() since it is inside it.

src/main/java/org/fenixedu/academic/domain/candidacy/FirstTimeDocumentsConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+        if (jsonElement != null) {
+            return Boolean.valueOf(jsonElement.getAsString());
```
sergiofbsilva: json supports boolean values, so this could be just `jsonElement.getAsBoolean()`

src/main/java/org/fenixedu/academic/domain/candidacy/FirstTimeDocumentsConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    public JsonElement getConfigurationProperties() {
+        return getJsonElement(super.getConfigurationProperties());
+    }
+
+    public JsonElement getJsonElement(JsonElement jsonElement) {
```
sergiofbsilva: this method should be private

src/main/dml/fenixedu-academic.dml
sergiofbsilva: The getter is overridden, however you should make this slot protected.

PR #1005, opened 2016-08-10
Make gratuity debts transferable to external entities ACDM-1108 #resolve

gil-l:
Allow consultation and liquidation of the new external entities' debts
Allow the creation, configuration and deletion of transfered debts' posting rules
Fix missing messages in some DomainExceptions in accounting

Debt transfer requires the external entity to be a member of a list of authorized
scholarship granting entities and a PDF document as proof of the entities acceptance
of the student's debt.
The new external entity's debt can be consulted and liquidated via the beneficiary
student's payment management area.
Entities may be added to the scholarship authorized list in the administrator tab.


Luis-Cruz:
After speaking with Gil, my second observation regarding the selection of 3rd party NIFs, can be accomplished by setting an already existing relation with a set of Parties.

PR #1017, opened 2016-09-15
Add the ability to fill the Identification Card's check digit ACDM-1055 #resolves

fernandojps:


fernandojps:
I made some changes to use the existing domain. Also, I edited the schema used in the person details editor, to add the check digit field. Should I edit the interface jsp instead?

fernandojps:
I've reused the Bean, but added a new schema.
When an update happens from ExtraDigit to SeriesNumber the SeriesNumber will be shown, even though both are stored in the database.

fernandojps:
It is now applied to all interfaces that use the schema student.documentId-edit.
The  interfaces in view mode will have the new field hidden, and the interfaces in edit mode will have the new field disabled if the type is not "IDENTITY_CARD".

sergiofbsilva:
This PR should also fix all the other interfaces that view/edit personal details, including manage accounts in account management.

PR #1024, opened 2016-10-17
Fix invalid dates returned by NextPossibleSummaryDate. ACDM-1134 #resolve

Paulo-Branco:


src/main/java/org/fenixedu/academic/domain/Lesson.java
```diff
@@ -691,7 +691,10 @@ public YearMonthDay getNextPossibleSummaryDate() {
             }
 
         } else {
-            YearMonthDay nextPossibleDate = hasAnyLessonInstances() ? getFirstLessonInstance().getDay() : getLessonStartDay();
+            SortedSet<YearMonthDay> validLessonDates = getAllValidLessonDatesWithoutInstancesDates(getLessonStartDay(), getLessonEndDay());
+            YearMonthDay nextValidDate = validLessonDates.size() > 0 ? validLessonDates.first() : null;
```
sergiofbsilva: This code shouldn't run if there are lesson instances

src/main/java/org/fenixedu/academic/domain/Lesson.java
```diff
@@ -691,7 +691,10 @@ public YearMonthDay getNextPossibleSummaryDate() {
...
+
+            YearMonthDay nextPossibleDate = hasAnyLessonInstances() ? getFirstLessonInstance().getDay() : nextValidDate;
```
sergiofbsilva: turn this into a if with the code above.

PR #1029, opened 2016-10-24
Create group of active phd students FIST-369 #resolves

fernandojps:


sergiofbsilva:
Since phd domain is to be refactored, lets not do this here but in the ist project.

PR #1033, opened 2016-10-31
Add signal when a professorship is created and deleted ACDM-1144 #resolve

diutsu:


src/main/java/org/fenixedu/academic/domain/Professorship.java
```diff
@@ -88,6 +94,7 @@ public void delete() {
         ProfessorshipManagementLog.createLog(getExecutionCourse(), Bundle.MESSAGING, "log.executionCourse.professorship.removed",
                 getPerson().getPresentationName(), getExecutionCourse().getNome(), getExecutionCourse()
                         .getDegreePresentationString());
+        ExecutionCourse executionCourse = getExecutionCourse();
```
Luis-Cruz: What is this variable for? It is never used in the remainder of the code block.

src/main/java/org/fenixedu/academic/domain/Professorship.java
```diff
@@ -45,6 +47,9 @@
     public static final Comparator<Professorship> COMPARATOR_BY_PERSON_NAME = new BeanComparator("person.name",
             Collator.getInstance());
 
+    public static final String PROFESSORSHIP_CHANGED = "academic.professorship.change";
```
Luis-Cruz: Why changed? The Professorship seem to be new.

src/main/java/org/fenixedu/academic/domain/Professorship.java
```diff
@@ -45,6 +47,9 @@
...
+    public static final String PROFESSORSHIP_DELETED = "academic.professorship.deleted";
```
Luis-Cruz: This is declared but then it is never used.
Are we missing a Signal.emit() somewhere, or is the constant no necessary?

src/main/java/org/fenixedu/academic/domain/Professorship.java
diutsu: The idea here was to trigger a signal on deletion, but currently that is not possible (the object will not exist on signal handler invocation). Pull request updated to remove these lines.

PR #1039, opened 2016-11-21
Country list now appears since EL is now ignored. ACDM-1145 #resolve

sergiofbsilva:


sergiofbsilva:
This is fixed with #1041

PR #1040, opened 2016-11-22
Use Program Conclusion for Alumni registration checks ACDM-1094 #resolve

Paulo-Branco:


src/main/java/org/fenixedu/academic/domain/AlumniManager.java
```diff
@@ -102,7 +103,8 @@ private void checkRulesToRegisterAlumni(final Integer studentNumber, final Strin
             if (!student.getRegistrationsSet().isEmpty()) {
 
                 for (Registration registration : student.getRegistrationsSet()) {
-                    if (registration.isConcluded()) {
+                    ProgramConclusion programConclusion = ProgramConclusion.conclusionsFor(registration).findAny().orElse(null);
+                    if (registration.isConcluded() || programConclusion != null) {
```
sergiofbsilva: use `Optional.isPresent` instead of checking nulls.

PR #1070, opened 2017-01-12
The display name is now visible and editable. ACDM-1186 #resolve

acfsilva:
The display name is now visible and editable on the Account Management interface.


src/main/java/org/fenixedu/academic/domain/Person.java
```diff
@@ -219,6 +219,11 @@ public void setGivenNames(String newGivenNames) {
         profile.changeName(newGivenNames, profile.getFamilyNames(), profile.getDisplayName());
     }
 
+    public void setNickname(String newNickname) {
```
sergiofbsilva: This changes the displayName of the UserProfile. Shouldn't the setter have a similar name ?

src/main/resources/resources/ApplicationResources_en.properties
```diff
@@ -2475,6 +2475,7 @@ label.extra.lesson = Extraordinary lesson
 label.false = No
 label.falseLabel = -
 label.familyNames = Family Names
+label.nickName = Display Name
```
sergiofbsilva: change accordingly.

src/main/resources/resources/ApplicationResources_pt.properties
```diff
@@ -2476,6 +2476,7 @@ label.extra.lesson = Aula Extraordin
 label.false = N�o
 label.falseLabel = -
 label.familyNames = Apelidos
+label.nickName = Nome que aparece na P�g. Pessoal
```
sergiofbsilva: Maybe 'Nome de Apresentação' ?

PR #1097, opened 2017-02-07
Optimized code. ACDM-1210 #resolve

Luis-Cruz:


src/main/java/org/fenixedu/academic/domain/Holiday.java
```diff
@@ -56,4 +57,14 @@ public static boolean isHoliday(LocalDate date, Space campus) {
...
     }
 
+    public static boolean isHoliday(YearMonthDay date, Space campus) {
+        for (Holiday holiday : Bennu.getInstance().getHolidaysSet()) {
+            if ((holiday.getLocality() == null || (campus != null && holiday.getLocality() == campus.getLocality()))
+                    && holiday.getDate().isMatch(date)) {
+                return true;
+            }
+        }
+        return false;
```
sergiofbsilva: why not stream anyMatch ?

Luis-Cruz: This code is called alot... in very intensive calculations... the overhead of creating a stream would degrade performance significantly, but I agree the code would look nicer.

PR #1113, opened 2017-02-16
Add valid id document checker to student warnings service. ACDM-1220 #resolve

pasadinhas:


src/main/java/org/fenixedu/academic/service/StudentWarningsService.java
```diff
@@ -14,6 +22,29 @@
 public class StudentWarningsService {
     private static List<Function<Student, Collection<String>>> checkers = new CopyOnWriteArrayList<>();
 
+    public final static Function<Student, Collection<String>> WARNING_VALID_ID_DOCUMENT = student -> {
```
sergiofbsilva: This should be invoked in [FenixInitializer](https://github.com/FenixEdu/fenixedu-academic/blob/v5.8.0/src/main/java/org/fenixedu/academic/servlet/FenixInitializer.java) and the code should be moved to a different new class where all the default checkers should reside.

pasadinhas:
New PR: https://github.com/FenixEdu/fenixedu-academic/pull/1118

PR #1115, opened 2017-02-20
Removed IST dependencies from PhoneUtil; Added new PhoneNumberHandler Interface ACDM-1234 #resolve

eralmeida:


src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,119 @@
...
+
+    static {
+        DEFAULT_RESOLVERS.put(EmailAddress.class, (pc) -> ((EmailAddress) pc).getValue());
+        DEFAULT_RESOLVERS.put(MobilePhone.class, (pc) -> ((MobilePhone) pc).getNumber());
+        DEFAULT_RESOLVERS.put(Phone.class, (pc) -> ((Phone) pc).getNumber());
+        DEFAULT_RESOLVERS.put(PhysicalAddress.class, (pc) -> ((PhysicalAddress) pc).getAddress());
+        DEFAULT_RESOLVERS.put(WebAddress.class, (pc) -> ((WebAddress) pc).getUrl());
+    }
+
+    private static final Collection<PhoneNumberType> FIXED_NUMBERS;
```
Luis-Cruz: You can init the constant in the same line.
`private static final Collection<PhoneNumberType> FIXED_NUMBERS = new ArrayList<>();`

src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final Collection<PhoneNumberType> MOBILE_NUMBERS;
```
Luis-Cruz: Here you can also init the constant right away.

src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,119 @@
...
+
+    static {
+        FIXED_NUMBERS = new ArrayList<PhoneNumberType>();
```
Luis-Cruz: If the constant is already initialised then this line can be deleted.

src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,119 @@
...
+        FIXED_NUMBERS.add(PhoneNumberType.VOIP);
+        FIXED_NUMBERS.add(PhoneNumberType.FIXED_LINE);
+
+        MOBILE_NUMBERS = new ArrayList<PhoneNumberType>();
```
Luis-Cruz: If the constant is already initialised then this line can be deleted.

src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,119 @@
...
+        MOBILE_NUMBERS.add(PhoneNumberType.MOBILE);
```
Luis-Cruz: You can use something like:
`Collections.singletonList(o)`
to init the list with a single element on creation.
This means this line can also be deleted.

src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,119 @@
...
+                numberText = numberText.replaceFirst("00", "+");
+            }
+
+            try {
+                final PhoneNumber phoneNumber = PHONE_UTIL.parse(numberText, null);
+                if (PHONE_UTIL.isValidNumber(phoneNumber)) {
+                    return phoneNumber;
+                }
+            } catch (NumberParseException e) {
+                System.out.println("O n�mero n�o � v�lido:" + e);
```
Luis-Cruz: Using System.out is strictly forbidden.
This line should be deleted.

src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,119 @@
...
+                return null;
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public String getInternationalFormatNumber(String numberText) {
+        final PhoneNumber phoneNumber = parsePhoneNumber(numberText);
+        if (phoneNumber != null) {
```
Luis-Cruz: Using the ternary operator here will result in nicer code:
`return phoneNumber != null ? PHONE_UTIL.format(phoneNumber, PhoneNumberFormat.E164) : null`

src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,119 @@
...
+            return PHONE_UTIL.format(phoneNumber, PhoneNumberFormat.E164);
+        }
+
+        return null;
+    }
+
+    @Override
+    public String resolve(PartyContact contact) {
+        return getResolver(contact.getClass()).getPresentationValue(contact);
```
Luis-Cruz: Calling `getResolver(contact.getClass())` may result in a null return value.
This should be checked because we never really know what the underlying map contains.

src/main/java/org/fenixedu/academic/util/PhoneUtil.java
```diff
@@ -18,134 +18,75 @@
...
     }
 
-    public static boolean isMobileNumber(String numberText) {
-        return isType(getPhoneNumberType(getPhoneNumber(numberText)), MOBILE_NUMBERS);
+    public static boolean isFixedNumber(String numberText) {
+        return PHONE_NUMBER_HANDLER.isFixedNumber(numberText);
     }
 
-    public static boolean isPortugueseNumber(String numberText) {
+    @Deprecated
```
Luis-Cruz: Why Deprecated? Is it never used? In that case it should be stated.
Is there an alternative method that should be used instead?

sergiofbsilva:
Due to JIRA smart commits, the commit must have ISSUE_KEY #resolve at the end of the commit message in order to be automatically resolved.

eralmeida:
Created new Pull Request ACDM-1234 for this issue

PR #1118, opened 2017-02-21
Add valid id document default checker. ACDM-1220 #resolve

pasadinhas:


src/main/java/org/fenixedu/academic/service/StudentWarningsDefaultCheckers.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.function.Function;
+
+public class StudentWarningsDefaultCheckers {
+
+    public final static Function<Student, Collection<String>> WARNING_VALID_ID_DOCUMENT = student -> {
+        Collection<String> warnings = new ArrayList<>();
+        Person person = student.getPerson();
+        if (person.getIdDocumentType().equals(IDDocumentType.IDENTITY_CARD)) {
```
Luis-Cruz: We can do the following instead, because enums are the same instance in the JVM
`person.getIdDocumentType() == IDDocumentType.IDENTITY_CARD`
It will have a slightly better performance, and this code will be called allot.

src/main/java/org/fenixedu/academic/service/StudentWarningsDefaultCheckers.java
```diff
@@ -0,0 +1,40 @@
...
+            if (person.getIdentificationDocumentSeriesNumber().length() == 1) {
```
Luis-Cruz: `person.getIdentificationDocumentSeriesNumber()`
should be attributed to a variable and reused.
This will avoid the code from inside this method running twice.

pasadinhas:
Changed based on review.

PR #1120, opened 2017-02-22
Separate active students group in two. ACDM-1195 #resolve

sergiofbsilva:


src/main/java/org/fenixedu/academic/domain/accessControl/ActiveStudentsGroup.java
```diff
@@ -63,9 +59,8 @@ public boolean isMember(User user) {
         if (user == null || user.getPerson() == null) {
             return false;
         }
-        return (user.getPerson().getStudent() != null && user.getPerson().getStudent().hasActiveRegistrations())
-                || user.getPerson().getPhdIndividualProgramProcessesSet().stream()
-                        .anyMatch(PhdIndividualProgramProcess::isProcessActive);
+        return new ActivePhdProcessesGroup().isMember(user) || (user.getPerson().getStudent() != null && user.getPerson()
```
Luis-Cruz: There are more non PHD students than there are PHD students, thus switching the order of the || will probably improve performance.

PR #1125, opened 2017-03-06
Removed IST dependencies from PhoneUtil; Added new PhoneNumberHandler Interface ACDM-1234 #resolve

eralmeida:


src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,112 @@
...
+import com.google.i18n.phonenumbers.PhoneNumberUtil;
+import com.google.i18n.phonenumbers.PhoneNumberUtil.PhoneNumberFormat;
+import com.google.i18n.phonenumbers.PhoneNumberUtil.PhoneNumberType;
+import com.google.i18n.phonenumbers.Phonenumber.PhoneNumber;
+
+public class AcademicPhoneNumberHandler implements PhoneNumberHandler {
+
+    protected static final PhoneNumberUtil PHONE_UTIL = PhoneNumberUtil.getInstance();
+
+    private static final Map<Class<? extends PartyContact>, ContactResolver> DEFAULT_RESOLVERS = new HashMap<>();
```
Luis-Cruz: How can another module add or replace a ContactResolver?
Don't we need an extra method in this class to register a ContactResolver?

src/main/java/org/fenixedu/academic/domain/contacts/PartyContact.java
```diff
@@ -176,7 +177,9 @@ public void edit(final PartyContactType type, final boolean defaultContact) {
         setDefaultContactInformation(defaultContact);
     }
 
-    public abstract String getPresentationValue();
+    public String getPresentationValue() {
+        return PhoneUtil.resolve(this);
```
Luis-Cruz: Why make PartyContact depend on PhoneUtil?
It's a very strange dependency. The PartyContact class should not have to know anything about PhoneUtil.

src/main/java/org/fenixedu/academic/util/AcademicPhoneNumberHandler.java
```diff
@@ -0,0 +1,112 @@
...
+import org.fenixedu.academic.domain.contacts.WebAddress;
+import org.fenixedu.academic.util.PhoneUtil.PhoneNumberHandler;
+
+import com.google.i18n.phonenumbers.NumberParseException;
+import com.google.i18n.phonenumbers.PhoneNumberUtil;
+import com.google.i18n.phonenumbers.PhoneNumberUtil.PhoneNumberFormat;
+import com.google.i18n.phonenumbers.PhoneNumberUtil.PhoneNumberType;
+import com.google.i18n.phonenumbers.Phonenumber.PhoneNumber;
+
+public class AcademicPhoneNumberHandler implements PhoneNumberHandler {
```
Luis-Cruz: This class is named AcademicPhoneNumberHandler but it contains references to other contact types such as EmailAddress, WebAddress, ... this makes no sense.

PR #1126, opened 2017-03-06
Check if username is valid ACDM-1231 #resolve

luisfsantos:


src/main/webapp/WEB-INF/resources/ApplicationResources_en.properties
```diff
@@ -139,6 +139,7 @@ label.title.StatuteTypeManagement = Manage Student Statutes
 label.true = Yes
 label.update=Update
 label.username = Username
+label.valid.istid = The username has to be valid.
```
sergiofbsilva: label.valid.username

src/main/webapp/WEB-INF/resources/ApplicationResources_pt.properties
```diff
@@ -139,6 +139,7 @@ label.title.StatuteTypeManagement = Gerir Estatutos de Aluno
 label.true = Sim
 label.update=Actualizar
 label.username = Username
+label.valid.istid = O utilizador tem de ser válido.
```
sergiofbsilva: label.valid.username

PR #1132, opened 2017-03-13
Removed RoleType dependency and added username filter ACDM-670 #resolve

eralmeida:


src/main/java/org/fenixedu/academic/dto/photographs/PhotographFilterBean.java
```diff
@@ -130,7 +142,7 @@ public boolean accepts(Photograph photograph) {
             if (getPersonType() != null && !getPersonType().isMember(person.getUser())) {
                 return false;
             }
-            if (getName() != null && !HumanName.namesMatch(person.getName(), name)) {
+            if (!getUserName().equals("") && !getUserName().equals(person.getUsername())) {
```
sergiofbsilva: better to use Strings.isNullOrEmpty

src/main/webapp/WEB-INF/fenixedu-academic/schemas/manager-schemas.xml
```diff
@@ -522,18 +522,8 @@
...
-	<slot name="personType" key="label.RoleType.name">
-        <property name="includedValues" value="ALUMNI, TEACHER, STUDENT, EMPLOYEE, GRANT_OWNER" />
-        <property name="key" value="true" />
-        <property name="defaultText" value="label.dropDown.all" />
-        <property name="bundle" value="ENUMERATION_RESOURCES" />
-    </slot>
-	<slot name="name" key="label.Party.name">
-	    <property name="size" value="50" />
+	<slot name="userName" key="label.Party.username">
+	    <property name="size" value="15" />
```
sergiofbsilva: Maybe this schema could be inlined in the jsp if not used elsewhere.

PR #1151, opened 2017-04-19
Don't catch domain exceptions during transactions ACDM-1279 #resolve

jrodriguesnum1:
Move the transaction code to a separate method when a domain exception was being caught during a transaction.


Luis-Cruz:
What Jira issue is this PullRequest related to?

The issue should be referenced in the header of the PullRequest.

Please refer to the following page for more information: https://confluence.fenixedu.org/display/FENIXEDU/Contributing

PR #1152, opened 2017-04-20
Evaluation Seasons added Fenix Bootstrap ACDM-1264 #resolve

eralmeida:


src/main/java/org/fenixedu/academic/bootstrap/FenixBootstrapper.java
```diff
@@ -154,6 +156,29 @@
         return Lists.newArrayList();
     }
 
+    private static void createEvaluationSeasons() {
```
Luis-Cruz: Creating an auxiliary method would greatly simplify the implementation of this method making it much easier to read.
I suggest something like this:
`private static void createEvaluationSeasons(final String enCode, final String ptCode, final String nameKey, final boolean normal, final boolean improvement, final boolean specialAuthorization, final boolean special) {
        final LocalizedString acronym = new LocalizedString.Builder().with(Locale.forLanguageTag("pt-PT"), enCode) .with(Locale.forLanguageTag("en-GB"), ptCode).build();
        new EvaluationSeason(acronym, BundleUtil.getLocalizedString(Bundle.ENUMERATION, nameKey), normal, improvement, specialAuthorization, special);        
}
`

src/main/java/org/fenixedu/academic/bootstrap/FenixBootstrapper.java
```diff
@@ -154,6 +156,29 @@
...
+
+        LocalizedString acronym = new LocalizedString.Builder().with(Locale.forLanguageTag("pt-PT"), "AE")
+                .with(Locale.forLanguageTag("en-GB"), "SA").build();
+        new EvaluationSeason(acronym, BundleUtil.getLocalizedString(Bundle.ENUMERATION, "SPECIAL_AUTHORIZATION"), false, false,
```
Luis-Cruz: With the previous auxillay method, these lines would be replaced by something like: 
`createEvaluationSeasons("AE", "SA", "SPECIAL_AUTHORIZATION", false, false, true, false);`

PR #1155, opened 2017-04-27
Added ECTS grade scale to student's curriculum view + converted grade column ACDM-727 #resolve

eralmeida:


src/main/java/org/fenixedu/academic/ui/struts/action/administrativeOffice/student/RegistrationDA.java
```diff
@@ -121,6 +124,14 @@ public ActionForward chooseCycleForViewRegistrationCurriculum(ActionMapping mapp
         final Person studentPerson = registrationCurriculumBean.getStudent().getPerson();
         request.setAttribute("studentPerson", studentPerson);
 
+        final EctsGraduationGradeConversionTable ectsGraduationGradeConversionTable =
+                EctsTableIndex.getGraduationGradeConversionTable(registrationCurriculumBean.getRegistration().getDegree(),
+                        registrationCurriculumBean.getRegistration().getCycleType(ExecutionYear.readCurrentExecutionYear()),
+                        ExecutionYear.readCurrentExecutionYear().getAcademicInterval(), new DateTime());
+        request.setAttribute("ectsGradeConversionTable", ectsGraduationGradeConversionTable.getEctsTable().toString().split(""));
```
Luis-Cruz: Why isn't this attribute and the next also added to the viewRegistrationCurriculum method of the same class?
Both methods forward to the same JSP, but only this method provides these attributes. Will the rendering of the JSP fail when forwarded to by the viewRegistrationCurriculum method?

PR #1160, opened 2017-05-10
Added CSV file extension to response header ACDM-1287 #resolve

acfsilva:


Luis-Cruz:
Reworked code: https://github.com/FenixEdu/fenixedu-academic/pull/1161

PR #1173, opened 2017-07-04
Add department selection for competence courses ACDM-1288 #resolve

luisfsantos:


src/main/webapp/bolonhaManager/competenceCourseVersions/listCompetenceCourses.jsp
```diff
@@ -27,7 +27,22 @@
 <%@page import="org.fenixedu.academic.ui.struts.action.BolonhaManager.CompetenceCourseInformationRequestBean"%><html:xhtml/>
 <%@page import="org.fenixedu.academic.domain.ExecutionSemester"%><html:xhtml/>
 
-	<h2><bean:write name="department" property="realName"/></h2>
+<h2><bean:write name="department" property="realName"/></h2>
```
Luis-Cruz: What happens if the department is null ?
I know this is not a new problem, the code before these changes also had this problem, but since we are making changes we can fix it also.

PR #1180, opened 2017-07-12
Centralize competence course management ACDM-1311 #resolve

sergiofbsilva:


Luis-Cruz:
I can't find an issue with reference ACDM-1331

PR #1213, opened 2017-09-18
Fix RuntimeException when adding an invalid authorization. ACDM-1278 #resolve

francisconeves97:


src/main/java/org/fenixedu/academic/ui/struts/action/academicAdministration/AuthorizationsManagementBean.java
```diff
@@ -128,7 +128,9 @@ public void editAuthorization(String parameter) {
     }
 
     public void addNewAuthorization() {
-        groups.add(0, new AuthorizationGroupBean());
+        if (groups.get(0).getRule() != null) {
```
sergiofbsilva: What if groups is empty at this point ?

PR #1218, opened 2017-10-03
Fix excel report for dissertations. ACDM-1315 #resolve

JCAbrantes:
Fix NPE (this report requires thesis to have an enrollment).
Remove 2 columns (these columns weren't adding new information).
Change delimiter character in "Affiliacao" and "Distribuicao Creditos" columns for better understanding.
Add column about first time enrollment.


sergiofbsilva:
Missing <issue> #resolve tag

PR #1225, opened 2017-10-23
WriteStudentAttendingCourse throws an exception on duplicate ExecutionCourse ACDM-1349 #resolve

JCAbrantes:
WriteAttendingCourse.run now throws an exception when the execution course is connected to other registration (the student is already enrolled in the course through other registration).

### DIFF
--- a/src/main/java/org/fenixedu/academic/service/services/enrollment/shift/WriteStudentAttendingCourse.java
+++ b/src/main/java/org/fenixedu/academic/service/services/enrollment/shift/WriteStudentAttendingCourse.java
@@ -18,6 +18,9 @@
  */
 package org.fenixedu.academic.service.services.enrollment.shift;
 
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import org.fenixedu.academic.domain.ExecutionCourse;
 import org.fenixedu.academic.domain.ExecutionSemester;
 import org.fenixedu.academic.domain.student.Registration;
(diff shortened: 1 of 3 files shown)


src/main/java/org/fenixedu/academic/service/services/enrollment/shift/WriteStudentAttendingCourse.java
```diff
@@ -40,7 +40,16 @@ protected void run(Registration registration, String executionCourseId) throws F
         if (registration == null) {
             throw new FenixServiceException("error.registration.not.exist");
         }
-        registration.addAttendsTo(readExecutionCourse(executionCourseId));
+
+        ExecutionCourse executionCourse = readExecutionCourse(executionCourseId);
+
+        for (Registration registration2 : registration.getStudent().getAllRegistrations()) {
+            if (!registration.getExternalId().equals(registration2.getExternalId()) && registration2.attends(executionCourse)) {
```
Luis-Cruz: The fenix-framework guarantees that each instance of a domain object is unique within the virtual machine. Therefore no need exists to compare the objects external id. Just compare the objects instead. The resulting code is easier to read and more efficient.

src/main/java/org/fenixedu/academic/service/services/enrollment/shift/WriteStudentAttendingCourse.java
```diff
@@ -40,7 +40,16 @@ protected void run(Registration registration, String executionCourseId) throws F
         if (registration == null) {
             throw new FenixServiceException("error.registration.not.exist");
         }
-        registration.addAttendsTo(readExecutionCourse(executionCourseId));
+
+        ExecutionCourse executionCourse = readExecutionCourse(executionCourseId);
+
+        for (Registration registration2 : registration.getStudent().getAllRegistrations()) {
```
Luis-Cruz: The method getAllRegistrations() is very heavy from a performance point-of-view.
If you look at the implementation of this method you will find that it constructs an unmodifiable collection based on an unmodifiable set that was constructed based on a new hash set that was filled with a filtered group of registrations. That's allot of unnecessary overhead. I suggest using the getRegistrationStream() method instead.

src/main/java/org/fenixedu/academic/service/services/enrollment/shift/WriteStudentAttendingCourse.java
```diff
@@ -40,7 +42,16 @@ protected void run(Registration registration, String executionCourseId) throws F
         if (registration == null) {
             throw new FenixServiceException("error.registration.not.exist");
         }
-        registration.addAttendsTo(readExecutionCourse(executionCourseId));
+
+        ExecutionCourse executionCourse = readExecutionCourse(executionCourseId);
+
+        for (Registration registration2 : registration.getStudent().getRegistrationsSet()){
```
Luis-Cruz: The getRegistrationsSet() method has the same problem of the getAllRegistrations() method.
Using getRegistrationStream() would be a better choice.

PR #1229, opened 2017-11-21
It should not be possible to create marksheets with annulled enrollment evaluations ACDM-1345 #resolve

danielcorreia96:
Changed message to describe the cause of the error properly to the user.


danielcorreia96:
Based on further testing, this issue is associated with a specific situation of special season students. 

Given a special season student and a teacher that submits a marksheet where he graded this student, if the academic office cancels the special season enrolment and then tries to confirm the marksheet, the confirmation fails because the EnrolmentEvaluationState associated with the student is Annulled.

More details in the JIRA issue.

PR #1232, opened 2017-12-04
Check if person exists no longer strictly matches identification number ACDM-1364 #resolve

JCAbrantes:
Check person: if there is no person with identification number it searches for people with givenName and familyName.


src/main/java/org/fenixedu/academic/ui/struts/action/accounts/ManageAccountsDA.java
```diff
@@ -74,21 +74,17 @@ public ActionForward prepareCreatePerson(ActionMapping mapping, ActionForm form,
...
-            if (!Strings.isNullOrEmpty(bean.getDocumentIdNumber())) {
-                stream = stream.filter(p -> p.getDocumentIdNumber().equals(bean.getDocumentIdNumber()));
-            }
             results = stream.collect(Collectors.toSet());
-        } else if (!Strings.isNullOrEmpty(bean.getDocumentIdNumber())) {
-            results = Person.findPersonByDocumentID(bean.getDocumentIdNumber());
-        } else {
-            results = Collections.emptySet();
-        }
+        }
```
sergiofbsilva: The ``results`` collection should keep all the results (searching for ``documentIdNumber`` and ``names``)
With this code it will only search using names if there is no match for the ``documentIdNumber`` which is limitative.

PR #1236, opened 2017-12-06
Department editing and creation - NullPointerException ACDM-1272 #resolve

danielcorreia96:
**Changes summary**
- Changed header to "Edit Department" on editing page.
- Defined "Cancel" action to resolve NPE.
- Used Bennu's Group.parse method to accept an access-group expression in the username field.
- Handled DomainExceptions thrown when creating departments with empty mandatory fields.


src/main/java/org/fenixedu/academic/ui/struts/action/manager/ManageAssociatedObjects.java
```diff
@@ -634,7 +640,7 @@ public ActionForward editDepartment(ActionMapping mapping, ActionForm form, Http
             d.setName(bean.getName());
             d.setRealName(bean.getRealName());
             d.setRealNameEn(bean.getRealNameEn());
-            d.setCompetenceCourseMembersGroup(User.findByUsername(bean.getUsername()).groupOf());
+            d.setCompetenceCourseMembersGroup(Group.parse(bean.getUsername()));
```
sergiofbsilva: bean ``getUsername`` should be changed to something more semantic like ``membersGroupExpression``

danielcorreia96: Since getUsername method is being used in "Academic Office creation" to link a single user, the bean should just be extended with a membersGroupExpression attribute.

danielcorreia96:
Update: Extended bean with Group expression attribute for better semantic.

PR #1242, opened 2018-01-18
It is possible to create registration certificate when not enrolled ACDM-1282 #resolve

JCAbrantes:


src/main/java/org/fenixedu/academic/ui/renderers/providers/ExecutionYearsForRegistrationCertificateProvider.java
```diff
@@ -0,0 +1,42 @@
+/**
+ * Copyright © 2002 Instituto Superior Técnico
```
sergiofbsilva: change the copyright accordingly

PR #1248, opened 2018-02-08
Added the missing degree type to the list and the justification type and fixed country editing error ACDM-1189 #resolve

AngelaFrancisca:


sergiofbsilva:
Fixed in #1250

PR #1259, opened 2018-04-19
CoordinatorGroup presentation name ignores responsible flag ACDM-1391 #resolve

danielcorreia96:


src/main/resources/resources/GroupNameResources_en.properties
```diff
@@ -23,7 +23,9 @@ label.name.AlumniGroup = Alumni{0}
 label.name.CampusEmployeeGroup = Employees of campus {0}
 label.name.CandidateGroup = Candidates
 label.name.CerimonyInquiryGroup = Inquired persons of {0}
-label.name.CoordinatorGroup = Coordination{0}{1}
+label.name.CoordinatorGroup= {0}Coordination{1}{2}
```
sergiofbsilva: is this correct ?

src/main/resources/resources/GroupNameResources_pt.properties
```diff
@@ -23,7 +23,9 @@ label.name.AlumniGroup = Alumni{0}
 label.name.CampusEmployeeGroup = Funcion�rios do campus {0}
 label.name.CandidateGroup = Candidatos
 label.name.CerimonyInquiryGroup = Pessoas alvo do inqu�rito {0}
-label.name.CoordinatorGroup = Coordena��o{0}{1}
+label.name.CoordinatorGroup= Coordena��o{0}{1}{2}
+label.name.CoordinatorGroup.responsible.false= \u0020Adjunta
```
sergiofbsilva: Is this file latin1 ? it should be .. please remove \u0020 character.

src/main/resources/resources/GroupNameResources_en.properties
danielcorreia96: Yes, because {0} matches the responsibleQualifier parameter.
Which means, in this case (EN), we get Responsible Coordination/Assistant Coordination.

{1} and {2} match the degreeType and degree parameters (previously {0} and {1}).

src/main/resources/resources/GroupNameResources_pt.properties
danielcorreia96: It's ISO-8859. There are other occurences of \u0020 in this file which is why I used this for explicit whitespace.

PR #1297, opened 2018-07-31
queue jobs should be prioritized ACDM-1429 #resolve

danielcorreia96:
This is a draft PR since most of the changes were cleanups: if this gets approved i'll fix the commits accordingly.

The first commit (6344737) has the relevant change for GEP queue jobs to have low priority.

The second commit (fa8cc0a) has cleanups of boilerplate code that could be automatically replaced:
- Stream folding: collect, sum, count, anyMatch, findFirst, ...
- Remove some usages of Apache CollectionUtils (select, filter)
- Unnecessary explicit array creation and types
- Replace anonymous classes with lambdas


src/main/java/org/fenixedu/academic/task/JobQueueDispatcher.java
```diff
@@ -118,7 +115,7 @@ protected void fail(final QueueJob job, Throwable t) {
             StringWriter sw = new StringWriter();
             PrintWriter pw = new PrintWriter(sw);
             t.printStackTrace(pw);
-            String subject = "Job " + job.getClass().getName() + "failed 3 times";
+            String subject = String.format("Job %sfailed 3 times", job.getClass().getName());
```
sergiofbsilva: missing space in format string

danielcorreia96: updated

PR #1364, opened 2018-10-16
InsurancePR: Add number of days to due date ACDM-1506 #resolve

sergiofbsilva:
```
alter table `POSTING_RULE` add `NUMBER_OF_DAYS_TO_CALCULATE_DUE_DATE` int(11) DEFAULT NULL;
alter table `EVENT` add `DUE_DATE` datetime NULL default NULL;
update EVENT set DUE_DATE = DATE_ADD(WHEN_OCCURED, INTERVAL 60 DAY) where oid2cname(OID) = 'org.fenixedu.academic.domain.accounting.events.insurance.InsuranceEvent';
update POSTING_RULE set NUMBER_OF_DAYS_TO_CALCULATE_DUE_DATE = 60 where oid2cname(OID) = 'org.fenixedu.academic.domain.accounting.postingRules.InsurancePR';
```

### DIFF
--- a/src/main/dml/fenixedu-academic.dml
+++ b/src/main/dml/fenixedu-academic.dml
@@ -1348,6 +1348,7 @@ valueType org.fenixedu.academic.util.sibs.incomming.SibsIncommingPaymentFileDeta
                 }
 
             class accounting.events.insurance.InsuranceEvent extends accounting.events.AnnualEvent {
+                DateTime dueDate;
             }
 
         class accounting.events.EnrolmentEvent extends accounting.AcademicEvent {
@@ -1690,6 +1691,7 @@ valueType org.fenixedu.academic.util.sibs.incomming.SibsIncommingPaymentFileDeta
     }
 
         class accounting.postingRules.AdministrativeOfficeFeeAndInsurancePR extends accounting.PostingRule {
+
         }
 
         class accounting.postingRules.GenericSingleEntryTypePR extends accounting.PostingRule {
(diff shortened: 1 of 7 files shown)


src/main/java/org/fenixedu/academic/domain/accounting/events/insurance/InsuranceEvent.java
```diff
@@ -74,6 +78,14 @@ private InsuranceEvent() {
     public InsuranceEvent(Person person, ExecutionYear executionYear) {
         this();
         init(EventType.INSURANCE, person, executionYear);
+        setupDueDate();
+    }
+
+    private void setupDueDate() {
+        final PostingRule postingRule = getPostingRule();
+        if (postingRule instanceof InsurancePR) {
```
danielcorreia96: Should we log "anomaly" situations where the PostingRule for InsuranceEvents isn't an InsurancePR to avoid surprises later on?

src/main/dml/fenixedu-academic.dml
```diff
@@ -1348,6 +1348,7 @@ valueType org.fenixedu.academic.util.sibs.incomming.SibsIncommingPaymentFileDeta
                 }
 
             class accounting.events.insurance.InsuranceEvent extends accounting.events.AnnualEvent {
+                DateTime dueDate;
```
danielcorreia96: Do we really need this in the dml/database? 
Couldn't we just calculate the dueDate as needed using getWhenOccured and numberOfDaysToCalculateDueDate?

sergiofbsilva: we need it since if somebody changes the PR then the past will change which is not what we want.

src/main/java/org/fenixedu/academic/domain/accounting/events/insurance/InsuranceEvent.java
sergiofbsilva: that is a good idea btw.

PR #1390, opened 2018-11-05
Interface for registering refunds. ACDM-1535 #doesnotresolveyet

Luis-Cruz:


sergiofbsilva:
this is will be integrated in #1391

PR #1399, opened 2018-11-16
fix stack overflow on Enrollment in Courses ACDM-1543 #resolve

danielfmpires:


sergiofbsilva:
@danielfmpires there is an extra commit in this PR which is not necessary

PR #1428, opened 2018-12-18
implement equals and hashcode ACDM-1570 #resolve

sergiofbsilva:


src/main/java/org/fenixedu/academic/domain/accounting/DueDateAmountMap.java
```diff
@@ -142,4 +142,17 @@ public void clear() {
     private static <T> T throwImmutable() {
         throw new UnsupportedOperationException("This instance is immutable");
     }
+
+    @Override
+    public boolean equals(final Object obj) {
+        if (!(obj instanceof DueDateAmountMap)) {
```
Luis-Cruz: Why not just do this instead:
`return obj instanceof DueDateAmountMap && innerMap.equals(obj);`

PR #1432, opened 2018-12-19
Improve RAIDES Academic Unit auto complete provider ACDM-1350 #resolve

danielcorreia96:
Note that this provider stopped using the previously implemented indexing strategy.

### DIFF
--- a/src/main/java/org/fenixedu/academic/domain/organizationalStructure/UnitName.java
+++ b/src/main/java/org/fenixedu/academic/domain/organizationalStructure/UnitName.java
@@ -18,6 +18,7 @@
  */
 package org.fenixedu.academic.domain.organizationalStructure;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


src/main/java/org/fenixedu/academic/domain/organizationalStructure/UnitName.java
```diff
@@ -223,44 +210,13 @@ private static void addDifferentUnitNames(final int size, final UnitNameLimitedO
...
-            final UnitNamePart unitNamePart = UnitNamePart.find(nameParts[0]);
-            if (unitNamePart != null && nameParts.length == 1) {
-                unitNameLimitedOrderedSet.addAll(unitNamePart.getUnitNameSet());
-            } else {
-                final Set<UnitName> unitNames =
-                        unitNamePart == null ? Bennu.getInstance().getUnitNameSet() : unitNamePart.getUnitNameSet();
-                for (final UnitName unitName : unitNames) {
-                    final String normalizedUnitName = unitName.getName();
-                    if (containsAllExactWords(normalizedUnitName, nameParts)) {
-                        if (unitName.getUnit().getCode() != null) {
```
ricardorcr: This line here garantees that the units sent in the RAIDES context have a code, as they must, removing this will make the information not compatible with what is expected by DGES.
This filter must be applied when searching for school units that are going to be used for RAIDES reports

danielcorreia96: Ok but that requirement is already satisfied in the overriden add function of ExternalAcademicUnitNameLimitedOrderedSet: 

https://github.com/FenixEdu/fenixedu-academic/blob/7fc51aac48ac91e6af2d156374b96fdb58348cfa/src/main/java/org/fenixedu/academic/domain/organizationalStructure/UnitName.java#L128-L132

Since the && operator uses short-circuit evaluation and there is StringUtils.isEmpty validation, invalid units are never added to the set (done by the super call).

PR #1435, opened 2018-12-26
Ensure deposit is always possible ACDM-1572 #resolve

sergiofbsilva:


src/main/webapp/WEB-INF/resources/ApplicationResources.properties
```diff
@@ -161,3 +161,5 @@ accounting.event.details.excessRefund.date=Data Reembolso
 accounting.event.details.excessRefund.amount=Advance Total Amount
 label.details=Details
 accounting.event.details.excessRefund.payment=Utilizado como pagamento
+warning.event.overflow.deposit=
```
danielcorreia96: Missing resources values.

PR #1436, opened 2018-12-26
Confirm when leaving page ACDM-1547 #resolve

sergiofbsilva:
* Improve payment details (bigger font, better detail)

### DIFF
--- a/src/main/webapp/WEB-INF/fenixedu-academic/accounting/events-multiple-payments-confirm.jsp
+++ b/src/main/webapp/WEB-INF/fenixedu-academic/accounting/events-multiple-payments-confirm.jsp
@@ -29,6 +29,15 @@
 
 <script type="text/javascript">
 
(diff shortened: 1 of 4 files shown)


src/main/webapp/WEB-INF/fenixedu-academic/accounting/events-multiple-payments-confirm.jsp
```diff
@@ -29,6 +29,15 @@
 
 <script type="text/javascript">
 
+    window.onbeforeunload = function() {
+        // this message is not shown anymore when using the most recent browsers.
```
danielcorreia96: This message is only shown in Internet Explorer 🤣
https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Browser_compatibility

PR #1459, opened 2019-02-05
Improve generic applications user experience ACDM-733 #resolve

danielcorreia96:
**Changes Done**
~~1. Filter candidacy periods that have already ended and sort by start date.~~
1. Sorted list of candidacy periods by start date.
2. Converted email from properties file to MessageTemplate that can be further customized in runtime.
3. Fix some missing "getContent" in JSPs.


danielcorreia96:
@Luis-Cruz , I didn't know candidates weren't supposed to view the entire list of periods.

I've updated the commit such that the list is only sorted by start date (but not filtered).

PR #1478, opened 2019-02-25
Filter cancelled events from person events chart ACDM-1611 #resolve

danielcorreia96:


sergiofbsilva:
This is no longer necessary due to ACDM-1614

PR #1485, opened 2019-02-26
Exempt when cancelling ACDM-1614 #resolve

sergiofbsilva:


src/main/java/org/fenixedu/academic/domain/accounting/Event.java
```diff
@@ -673,39 +673,58 @@ public void forceCancel(final Person responsible, final String cancelJustificati
...
+                            EventExemptionJustificationType.CANCELLED, new DateTime(), justification);
+            fixedAmountInterestExemption.setWhenCreated(when);
+            when = when.plusSeconds(1);
         }
 
-        if (getPayedAmount().isPositive()) {
-            throw new DomainException("error.accounting.Event.cannot.cancel.events.with.payed.amount.greater.than.zero");
+        if (dueFineAmount.isPositive()) {
+            FixedAmountFineExemption fixedAmountFineExemption =
+                    new FixedAmountFineExemption(this, responsible, dueInterestAmount,
```
danielcorreia96: This should probably be `dueFineAmount` ⚠️

src/main/java/org/fenixedu/academic/domain/accounting/Event.java
```diff
@@ -673,39 +673,58 @@ public void forceCancel(final Person responsible, final String cancelJustificati
...
-        }
-        
-        if (!isOpen()) {
-            throw new DomainException("error.accounting.Event.only.open.events.can.be.cancelled");
+        DebtInterestCalculator debtInterestCalculator = getDebtInterestCalculator(when);
+        Money dueInterestAmount = new Money(debtInterestCalculator.getDueInterestAmount());
+        Money dueFineAmount = new Money(debtInterestCalculator.getDueFineAmount());
+        Money dueAmount = new Money(debtInterestCalculator.getDueAmount());
+
+        when = when.plusSeconds(1);
```
danielcorreia96: I understand why we need this type of construct (to exempt an event we need exempt the interests/fines first) but this usage probably isn't needed here. 

The `when.plusSeconds(1)`are already properly managed in the ifs.

src/main/java/org/fenixedu/academic/domain/accounting/Event.java
danielcorreia96:
```suggestion
                    new FixedAmountFineExemption(this, responsible, dueFineAmount,
```

PR #1487, opened 2019-02-27
Allow partial refund of the value payed for debt ACDM-1618 #resolve

ricardorcr:


src/main/java/org/fenixedu/academic/domain/accounting/Event.java
```diff
@@ -1147,52 +1150,39 @@ public boolean isOpenAndAfterDueDate(DateTime when) {
...
-        final BigDecimal dueInterestAmount = debtInterestCalculator.getDueInterestAmount();
-        boolean createdInterestOrFineExemptions = false;
-
-        if (dueInterestAmount.signum() == 1) {
-            FixedAmountInterestExemption fixedAmountInterestExemption =
-                    new FixedAmountInterestExemption(this, creator.getPerson(), new Money(dueInterestAmount), justificationType,
-                            now, reason);
-            fixedAmountInterestExemption.setWhenCreated(now.plusSeconds(1));
-            createdInterestOrFineExemptions = true;
+        final BigDecimal paidDebtAmount = debtInterestCalculator.getPaidDebtAmount().subtract(debtInterestCalculator.getTotalRefundAmount());
```
ricardorcr: this is because several refunds can be made, and as such we should subtract the amount already refunded from the amount payed

sergiofbsilva:
included in #1488

PR #1516, opened 2019-05-17
Create mobility program if mobility ACDM-1648 #resolve

sergiofbsilva:


src/main/java/org/fenixedu/academic/domain/student/RegistrationProtocol.java
```diff
@@ -128,6 +132,9 @@ public void edit(String code, LocalizedString description, Boolean enrolmentBySt
         setAlien(isAlien);
         setExempted(exempted);
         setMobility(mobility);
+        if (mobility && getMobilityProgram() == null) {
+            new MobilityProgram(this);
+        }
```
sergiofbsilva: @Luis-Cruz this should be enough, tough I am not sure if it is possible to delete the mobility program if mobility is put to false.

PR #1558, opened 2019-10-15
Add support for custom payment plans. ACDM-1687 #resolve

Luis-Cruz:


src/main/java/org/fenixedu/academic/domain/accounting/Event.java
```diff
@@ -653,7 +642,7 @@ public Money getTotalAmountToPay() {
     }
 
     public Money getOriginalAmountToPay() {
-        return new Money(getDebtInterestCalculator(getWhenOccured().plusSeconds(1)).getDebtAmount());
+        return new Money(getDebtInterestCalculator(new DateTime()).getDebtAmount());
```
sergiofbsilva: 😮😮😮 beware 😮😮😮

src/main/java/org/fenixedu/academic/domain/accounting/Event.java
```diff
@@ -18,26 +18,7 @@
...
-import java.util.Optional;
-import java.util.Set;
-import java.util.SortedSet;
-import java.util.TreeSet;
-import java.util.function.BiFunction;
-import java.util.function.Predicate;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
+import org.apache.solr.common.util.Hash;
```
sergiofbsilva: why are we using this import? 🤔

src/main/java/org/fenixedu/academic/domain/accounting/Event.java
Luis-Cruz: The getDebtAmount() originally only iterates the "DebtStream" and all "Debt" objects are always added to the calculator, irregardless of the instance in time being considered; in other words, as long as the sum of values of the inner DueDateAmountMap remains the same then the getDebtAmount() will always produce the same value for every instance in time.

Either way, we ran a script for every existing event to check, and of course the value always matched.

src/main/java/org/fenixedu/academic/domain/accounting/Event.java
Luis-Cruz: It was a mistake.

PR #1567, opened 2019-11-18
Add support for quarters/trimesters for curricular period creation ACDM-1694 #resolve

Luis-Cruz:


src/main/java/org/fenixedu/academic/util/CurricularPeriodLabelFormatter.java
```diff
@@ -33,30 +33,35 @@
...
 
-    public static String getFullLabelI18N(CurricularPeriod curricularPeriod, boolean abbreviated, Locale locale) {
+    public static String getFullLabelI18N(CurricularPeriod curricularPeriod, Integer term, boolean abbreviated, Locale locale) {
         StringBuilder result = new StringBuilder();
         while (curricularPeriod.getParent() != null) {
-            buildLabel(curricularPeriod, result, abbreviated, locale);
+            buildLabel(curricularPeriod, term, result, abbreviated, locale);
 
             curricularPeriod = curricularPeriod.getParent();
+            term = null;
```
ricardorcr: why are you setting "term" to null?

src/main/java/org/fenixedu/academic/ui/faces/bean/bolonhaManager/curricularPlans/CurricularCourseManagementBackingBean.java
```diff
@@ -21,29 +21,11 @@
...
-import org.fenixedu.academic.domain.CompetenceCourse;
-import org.fenixedu.academic.domain.CurricularCourse;
-import org.fenixedu.academic.domain.Degree;
-import org.fenixedu.academic.domain.DegreeCurricularPlan;
-import org.fenixedu.academic.domain.Department;
-import org.fenixedu.academic.domain.ExecutionDegree;
-import org.fenixedu.academic.domain.ExecutionSemester;
-import org.fenixedu.academic.domain.ExecutionYear;
-import org.fenixedu.academic.domain.Person;
-import org.fenixedu.academic.domain.Teacher;
```
ricardorcr: this imports should be the same instead of .*

src/main/java/org/fenixedu/academic/domain/degreeStructure/Context.java
```diff
@@ -134,6 +128,7 @@ public Context(final CourseGroup courseGroup, final DegreeModule degreeModule, f
         super.setParentCourseGroup(courseGroup);
         super.setChildDegreeModule(degreeModule);
         super.setCurricularPeriod(curricularPeriod);
+        super.setTerm(term == null || term.intValue() < 1 ? null : term);
```
ricardorcr: shouldn't this return an error message if term < 1?

src/main/java/org/fenixedu/academic/util/CurricularPeriodLabelFormatter.java
Luis-Cruz: It's in a loop.
If we don't clear it, after the first time it's printed then it will be repeated for each "level".

src/main/java/org/fenixedu/academic/domain/degreeStructure/Context.java
Luis-Cruz: No. The term is optional. If the context represents a year or a semester, then there is no term.

PR #1610, opened 2020-03-05
Not mandatory anymore mail in course page ACDM-1739 #resolve

ricardorcr:


ricardorcr:
The mail is not mandatory in the page, it's only in this interface.
Most courses do not have a mail associated to it, and once an email is entered is no longer possible to delete it.

PR #1732, opened 2021-04-22
Hide projects past end date inside a dropdown ACDM-1849 #resolve

diogotcorreia:
This PR makes past projects (partially) hidden from the student's view.
Some disciplines have huge amounts of projects (e.g. 5 per week) and it gets hard to find the relevant project.

### Progress

The initial idea was to add a toggle in the teacher's options.  
However, this was later changed to automatically separate past projects from upcoming/active projects.

A project is flagged as "past" when 1 day has passed since its end date.

It uses a standard `<details>` HTML tag along with a `<summary>` tag.

### Screenshots

image

### DIFF
--- a/src/main/webapp/student/projectSubmissions/viewProjectsWithOnlineSubmission.jsp
+++ b/src/main/webapp/student/projectSubmissions/viewProjectsWithOnlineSubmission.jsp
@@ -47,7 +47,7 @@
 			<ul>
 				<logic:iterate id="projectWithOnlineSubmission" name="projectsWithOnlineSubmission" type="org.fenixedu.academic.domain.Project">
 					<bean:define id="projectId" name="projectWithOnlineSubmission" property="externalId" />
-					<logic:equal value="false" name="projectWithOnlineSubmission" property="pastSubmission">
+					<logic:equal value="false" name="projectWithOnlineSubmission" property="pastProject">
 						<li>
 							<bean:write name="projectWithOnlineSubmission" property="name"/> ,
 							<html:link action="<%="/projectSubmission.do?method=viewProjectSubmissions&amp;attendsId=" + attendsId + "&amp;projectId=" + projectId%>">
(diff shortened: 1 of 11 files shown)


src/main/webapp/student/projectSubmissions/viewProjectsWithOnlineSubmission.jsp
```diff
@@ -47,12 +47,14 @@
 			<ul>
 				<logic:iterate id="projectWithOnlineSubmission" name="projectsWithOnlineSubmission" type="org.fenixedu.academic.domain.Project">
 					<bean:define id="projectId" name="projectWithOnlineSubmission" property="externalId" />
-					<li>
-						<bean:write name="projectWithOnlineSubmission" property="name"/> , 
-						<html:link action="<%="/projectSubmission.do?method=viewProjectSubmissions&amp;attendsId=" + attendsId + "&amp;projectId=" + projectId%>">
-							<bean:message key="link.projectSubmissions.viewProjectsWithOnlineSubmission.viewProjectSubmissions"/>
-						</html:link>
-					</li>
+					<logic:notEqual value="false" name="projectWithOnlineSubmission" property="visible">
```
sergiofbsilva, line 50: Double negatives are not that intuitive for who is reading :) You can read a little bit more about it here: https://refactoring.com/catalog/removeDoubleNegative.html
Do you think it is better this way?
```suggestion
					<logic:equal value="true" name="projectWithOnlineSubmission" property="visible">
```

diogotcorreia: I had that initially. However, it made so that the projects with the default value (`NULL`) were hidden, which would affect all existing projects.
Would there be a better way to migrate existing projects to have the `true` value by default?

Another option that I have thought, would be to instead rename this new property to `archive`, which might be more intuitive and solve the migration problem.

sergiofbsilva: I think you've fixed that in the `getVisible` method of the bean right? I believe it would be better to handle the null value by overriding the `getVisible` method in the domain class `Project`. The returned Boolean will never be null.
The migration is another possibility.
Nevertheless, this suggestion still applies, right? Or is there any quirk I am not recalling? 😅

diogotcorreia: Yes, I'll definitely look into that!  
Thanks a lot for the feedback :+1:

PR #1735, opened 2021-04-30
Fixed max and min credits check for OR rules ACDM-1842 #resolve

ricardorcr:


ricardorcr:
A new and more complete issue was created. This PR would not solve correctly the problem.

PR #1747, opened 2021-06-15
Add Extraordinary Evaluation Season - ACDM-1864 #resolve

diogotcorreia:
### Scripts

Add Extraordinary Season
```java
package org.fenixedu.academic.task;

import org.fenixedu.academic.domain.EvaluationSeason;
import org.fenixedu.academic.util.Bundle;
import org.fenixedu.bennu.core.i18n.BundleUtil;
import org.fenixedu.bennu.scheduler.custom.CustomTask;
import org.fenixedu.commons.i18n.LocalizedString;

import java.util.Locale;

public class CreateExtraordinarySeason extends CustomTask {

    public void runTask() {
        final LocalizedString acronym = new LocalizedString.Builder().with(Locale.forLanguageTag("pt-PT"), "EX")
                .with(Locale.forLanguageTag("en-GB"), "ES").build();
        new EvaluationSeason(acronym, BundleUtil.getLocalizedString(Bundle.ENUMERATION, "EXTRAORDINARY_SEASON"), false, false,
                false, false, true);
    }

}
```

Add Posting Rule
```java
package org.fenixedu.academic.task;

import org.fenixedu.bennu.scheduler.custom.CustomTask;
import org.fenixedu.academic.domain.accounting.ServiceAgreementTemplate;
import org.fenixedu.academic.domain.accounting.postingRules.ExtraordinarySeasonEnrolmentPR;
import pt.ist.fenixframework.FenixFramework;
import org.joda.time.DateTime;
import org.fenixedu.academic.util.Money;

import java.util.Locale;

public class CreateExtraordinarySeasonPR extends CustomTask {

    public void runTask() {
              ServiceAgreementTemplate sat = (ServiceAgreementTemplate) FenixFramework.getDomainObject("2383706849302");
               new ExtraordinarySeasonEnrolmentPR(new DateTime().minus(1000), null, sat, Money.valueOf(20));
    }

}
```

### Other Breaking Changes

This PR also renames the special season listing page in the administrative office section, therefore requiring it to be recreated in the menu.
- Academic Administration -> Lists -> Enrollments in Special Season

Since this updates strings in ApplicationResources, PR https://github.com/ist-dsi/fenix-webapp/pull/85 needs to be merged as well.

Furthermore, due to the renaming of some classes, leftover code on fenixedu-ist needs to be removed through PR https://github.com/ist-dsi/fenixedu-ist/pull/793


Luis-Cruz (changes requested):
Still missing report to allow services to identify enrolled studens
screenshot

Also, can't find entry in the price table to configure the rule. Does thiss need to be done via script?

diogotcorreia:
I've updated the top-level comment with the scripts

Luis-Cruz (changes requested):
Debts should not be checked.
[noDebtChecks.patch.txt](https://github.com/FenixEdu/fenixedu-academic/files/6770818/noDebtChecks.patch.txt)

Extraordinary exam season not available in resource managment portal:
screenshot3

Teachers should also be able to view students enrolled in extraordinary season:
screenshot2
screenshot1

Luis-Cruz (approved):

src/main/java/org/fenixedu/academic/domain/studentCurriculum/StudentCurricularPlanEnrolmentInExtraordinarySeasonEvaluationManager.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    @Override
+    protected void assertStudentEnrolmentPreConditions() {
+
+        if (!getRegistrationsToEnrolByStudent(getResponsiblePerson().getStudent()).contains(getRegistration())) {
+            throw new DomainException("error.StudentCurricularPlan.student.is.not.allowed.to.perform.enrol");
+        }
+
```
Luis-Cruz, line 75: A change in requirements… Similarly to the special season case, we need to explicitly check StatuteType to allow each student access to this evaluation season.

Here we neet to add that check. This means we will also need an extra flag in the StatuteType: extraordinarySeasonGranted

Luis-Cruz (changes requested):

src/main/java/org/fenixedu/academic/domain/student/StatuteType.java
```diff
@@ -141,6 +142,10 @@ public void delete() {
         return readAll(StatuteType::isSpecialSeasonGrantedByRequest).findFirst();
     }
 
+    public static Optional<StatuteType> findExtraordinarySeasonGrantedStatuteType() {
+        return readAll(StatuteType::isExtraordinarySeasonGranted).findFirst();
```
diogotcorreia, line 146: Would this be the best way to do this?
I didn't want to duplicate the "SpecialSeasonGrantedByRequest" since it was deprecated.
I feel like this might have unexpected behaviour if there was more than one statute with extraordinary season granted.

Luis-Cruz: I agree it's strange... and some code cleanup could avoid such solutions as findFirst or iterator.next...
In this case just to exactly that.
The purpose of the statutetype is to grant access to each evaluation season.
Some students indeed have multiple types that will grant them access, but which one we use has no real bearing on the result.

Luis-Cruz (approved):

PR #1807, opened 2021-10-28
Add callback to login button on 404 page | ACDM-1931 #resolve

diogotcorreia:
[ACDM-1931](https://jira.fenixedu.org/browse/ACDM-1931)


Luis-Cruz:
I think there are cases in which this will not work properly.
In some scenarios the URL is not enough no redirect the user (when other context is required to load the page that is lost with the session).
The other case is when the URL contains a checksum. In this case the callback will never work becasuse the checksum dependes on the session which is lost.

PR #1812, opened 2021-11-03
Reviewed isPartialRegime ACDM-1935 #resolve

ricardorcr:


src/main/java/org/fenixedu/academic/domain/student/Registration.java
```diff
@@ -3196,7 +3197,14 @@ public RegistrationRegimeType getRegimeType(final ExecutionYear executionYear) {
...
 
     public boolean isPartialRegime(final ExecutionYear executionYear) {
-        return getRegimeType(executionYear) == RegistrationRegimeType.PARTIAL_TIME;
+        final Optional<RegistrationDataByExecutionYear> data = getRegistrationDataByExecutionYearSet().stream()
+                .filter(rd -> rd.getExecutionYear() == executionYear)
+                .findAny();
+        if (data.isPresent()) {
+            return data.get().getMaxCreditsPerYear() != null;
+        } else {
+            return getRegimeType(executionYear) == RegistrationRegimeType.PARTIAL_TIME;
```
Luis-Cruz, line 3206: This case will only happen in years where there is no RegistrationDataByExecutionYear object.
Every past year will always return false.

Luis-Cruz (changes requested):

Luis-Cruz (approved):

